PR #7, opened 2019-12-09
run tests on all pushes, only deploy on master

gidsg:
We should run CI on all branches but only deploy on `master`


gidsg:
> It'd be good to update this at some point to only push to master if tests pass successfully, etc.

@stuartmccoll my understanding is that this is the default behaviour of GitHub actions, i.e. it will not run any following step if a preceding one fails. But we can review if we see this behaving differently in practice.

PR #10, opened 2019-12-09
filter empty lines from postcode select

gidsg:
Previously we were showing empty fields in the postcode select with commas.

Screenshot 2019-12-09 at 13 53 47
This filters empty fields out and refactors to a helper with associated test.


test/controllers/postcode_controller_test.rb
```diff
@@ -1,7 +1,12 @@
 require 'test_helper'
 
 class PostcodeControllerTest < ActionDispatch::IntegrationTest
-  # test "the truth" do
-  #   assert true
-  # end
+  include PostcodeHelper
```
stuartmccoll: Is this a Rails pattern, to test helpers inside the test for the controller? I noticed that there's a `helpers` directory under the `test` directory, which is where I'd have been inclined to put this in its own `postcode_helpers_test.rb` file, but happy for it to stay where it is if that's a standard pattern/best practice?

gidsg: No, I think you are right, I will move it!

PR #13, opened 2019-12-11
Feature/initial page stubs

PushBackNeil:


.gitignore
```diff
@@ -36,3 +36,7 @@ yarn-debug.log*
 # Credentials
 .env
 .DS_Store
+.generators
+.idea
+scaff.sh
+/nbproject/private/
```
gidsg: These look like personal things that I don't think should be in the gitignore. Can we remove them? You can add them to your global gitignore i.e. https://help.github.com/en/github/using-git/ignoring-files#create-a-global-gitignore

.rakeTasks
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
gidsg: I don't think this should be committed. I guess this is an autogenerated RubyMine file?

app/controllers/home_controller.rb
```diff
@@ -3,6 +3,6 @@ class HomeController < ApplicationController
   before_action :authenticate_user!
   
   def show
-    redirect_to '/dashboard'
+    #redirect_to '/dashboard'
```
gidsg: rather remove this line, rather than comment it out.

.gitignore
```diff
@@ -1,4 +1,4 @@
-# See https://help.github.com/articles/ignoring-files for more about ignoring files.
+# See https://help.github.com/articles/ignoring-files for more project_title ignoring files.
```
stuartmccoll: This looks like something picked up in a find and replace, which can be reverted 👍.

test/fixtures/organisations.yml
```diff
@@ -1,4 +1,4 @@
-# Read about fixtures at https://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html
+# Read project_title fixtures at https://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html
```
stuartmccoll: Looks like something else changed in a find and replace - there's a few of these in the `.yml` files in `test/fixtures/`, I won't add comments to them all.

.gitignore
gidsg: I guess this change is accidental, can we revert it?

config/routes.rb
```diff
@@ -1,5 +1,47 @@
...
+    get 'location' => 'project_location#project_location'
+    post 'save-location' => 'project_location#save_project_location'
+    get 'other-location' => 'project_location#project_other_location'
+    get 'key-dates' => 'project_dates#project_dates'
+    get 'title' => 'project_title#project_title'
+    get 'new-project' => 'new_project/new_project'
+  end
+
+  namespace :grant do
+    get 'application' => 'grant_application#grant_application'
```
gidsg: We don't really distinguish between the `project` and the `grant` as done here, so I would rather group all these under `project` if possible.

app/views/projects/new.html.erb
```diff
@@ -1,7 +1,7 @@
 <h1>New Project</h1>
 
 <%= form_with model: @project, class: "nifty_form" do |f| %>
-    <%= label_tag(:project_title, "Project title") %>
+    <%= label_tag(:project_title, "Project project_title") %>
```
stuartmccoll: Is this right? Doesn't matter too much if the page is still just a stub.

gidsg:
@PushBackNeil can you take a look at failing tests. You can run `rails test` locally to run these.

.gitignore
PushBackNeil: I have reverted this

test/fixtures/organisations.yml
PushBackNeil: Ive reverted these too

.rakeTasks
gidsg: @PushBackNeil I think you missed this but I will delete it in master.

PR #15, opened 2019-12-11
capital works content

PushBackNeil:


app/views/project/project_description/project_description.html.erb
```diff
@@ -18,7 +18,7 @@
   </form>
 
   <% if Rails.env.development? %>
```
gidsg: Can we add a `TODO` to remove these next links once the proper routing is done?

PushBackNeil: Will do in the next feature branch

PR #16, opened 2019-12-11
project permission content

PushBackNeil:


PushBackNeil:
@stuartmccoll I have added the TODO for the temporary links as requested

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_12_11_145241) do
```
gidsg: I don't think this change should be in there as you are not editing the database. Can you revert this?

gidsg: OK I think this is because the datestamp on the migration does not match the schema. I think it is safe then to commit it.

PR #22, opened 2019-12-12
project involvement content

PushBackNeil:


app/views/project/project_involvement/project_involvement.html.erb
```diff
@@ -1,2 +1,48 @@
...
+        <div class="govuk-inset-text">
+          This is a mandatory outcome for all projects funded by the National Lottery Heritage Fund.
+        </div>
+        <span id="with-hint-hint" class="govuk-hint">
+          Your answer should tell us how you will reach new groups of people through your projects. Include evidence to support your plans.
+        </span>
+        <textarea class="govuk-textarea govuk-js-character-count" id="project-involvement" name="project-involvement" rows="5" aria-describedby="project-involvement-info with-hint-hint"></textarea>
+      </div>
+      <span id="project-involvement-info" class="govuk-hint govuk-character-count__message" aria-live="polite">
+        You have 500 words remaining
```
gidsg: Does word count work yet, do we need some JS for this? If so maybe add a TODO?

PushBackNeil: The counters are not working yet, as the govuk js is not yet being used. I will add the javascript required in my next feature branch.

PR #24, opened 2019-12-12
Implemented first stage of wizard routing for organisations

stuartmccoll:


app/controllers/home_controller.rb
```diff
@@ -1,7 +1,30 @@
...
+
+    @organisation = Organisation.create
+
+    logger.debug 'Finished creating organisation ' + @organisation.id +
+                     ' for user ID: ' + current_user.id.to_s
+
+    logger.debug 'Updating user ID: ' + current_user.id.to_s +
+                     ' with link to organisation ' + @organisation.id
+
+    @user = User.update(current_user.id, :organisation_id => @organisation.id)
```
gidsg: Just a style thing but in modern ruby versions you can use `:` syntax instead of `=>` if the key is symbol safe e.g. `organisation_id: @organisation.id` should be equivalent.

app/controllers/organisation/organisation_mission_controller.rb
```diff
@@ -1,4 +1,41 @@
 class Organisation::OrganisationMissionController < ApplicationController
-  def organisation_mission
+  include Wicked::Wizard
+
+  steps :mission
```
gidsg: I am getting an error `undefined method `mission'` do you need to rebase to get the migration in this branch?

app/controllers/organisation/organisation_numbers_controller.rb
```diff
@@ -0,0 +1,42 @@
+class Organisation::OrganisationNumbersController < ApplicationController
+  include Wicked::Wizard
+
+  steps :numbers
+
+  before_action :authenticate_user!\
```
gidsg: think this `\` is a typo?

app/controllers/organisation/organisation_about_controller.rb
```diff
@@ -1,4 +1,42 @@
...
+  def update
+
+    @organisation = Organisation.find(current_user.organisation.id)
+
+    logger.debug 'Updating organisation ' + @organisation.id +
+                     ', setting name to ' + params[:organisation][:name] +
+                     ', postcode to ' + params[:organisation][:postcode] +
+                     ' and line1 to ' + params[:organisation][:line1]
+
+    @organisation = Organisation.update(current_user.organisation.id,
```
gidsg: If we add allowed params to the model to you can do a mass update of all params to reduce some of this code. See example: https://edgeguides.rubyonrails.org/action_controller_overview.html#strong-parameters

app/controllers/organisation/organisation_mission_controller.rb
stuartmccoll: Yep, the migration previously being reverted out caused this - it's back in now, so you shouldn't experience the same issue.

app/controllers/organisation/organisation_about_controller.rb
stuartmccoll: Much nicer - changed in the most recent commit.

app/controllers/organisation/organisation_numbers_controller.rb
stuartmccoll: Fixed 👍.

app/controllers/home_controller.rb
stuartmccoll: Fixed 👍.

Gemfile
```diff
@@ -71,5 +71,6 @@ gem "pry", "~> 0.12.2"
 
 gem "devise", "~> 4.7"
 
+gem "wicked", "~> 1.3.4"
```
gidsg: I think we should remove the `.4` here so it will always use the latest minor version.

app/controllers/organisation/organisation_mission_controller.rb
```diff
@@ -1,4 +1,41 @@
...
   end
+
+  def update
+
+    # This information about an organisation doesn't have to be provided,
+    # therefore we only want to update the model if the user has selected
+    # a form option. We can tell whether this has been done by checking
+    # for the presence of an organisation hash.
+    if !params[:organisation].nil?
+      @organisation = Organisation.find(current_user.organisation.id)
```
gidsg: This seems a bit weird to assign the same variable twice with different values. Could we find and update in one step?

app/controllers/organisation/organisation_numbers_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+    company_number = params[:organisation][:company_number].present? ? params[:organisation][:company_number] : nil
+    charity_number = params[:organisation][:charity_number].present? ? params[:organisation][:company_number] : nil
+
+    @organisation = Organisation.find(current_user.organisation.id)
+
+    logger.debug 'Updating organisation ' + @organisation.id +
+                     ', setting company_number to ' + company_number.to_s +
+                     ' and setting charity_number to ' + charity_number.to_s
+
+    @organisation = Organisation.update(current_user.organisation.id,
```
gidsg: We could use mass-assignment here?

app/controllers/organisation/organisation_numbers_controller.rb
```diff
@@ -0,0 +1,42 @@
+class Organisation::OrganisationNumbersController < ApplicationController
+  include Wicked::Wizard
+
+  steps :numbers
+
+  before_action :authenticate_user!
+
+  def show
+
+    @organisation = Organisation.find(current_user.organisation.id)
```
gidsg: We can DRY this by fetching the Organisation in a `before_action` e.g.
```
before_action :authenticate_user!, :set_organisation
```


```
  private
    def set_organisation
      @organisation = Organisation.find(current_user.organisation.id)
    end
```
Then `@organisation` is available globally across the class and we can call it directly e.g. `@organisation.update`

app/controllers/organisation/organisation_numbers_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    logger.debug 'Finished updating organisation ' + @organisation.id
+
+    redirect_to_finish_wizard
+
+  end
+
+  def finish_wizard_path
+
+    organisation_organisation_about_get_path(current_user.organisation.id, id:'about')
```
gidsg: We don't need the `.id` here. It will resolve it to the same value without it.

app/views/organisation/organisation_mission/mission.html.erb
```diff
@@ -0,0 +1,51 @@
...
+          <h1 class="govuk-fieldset__heading">
+            Tell us about the mission, or objectives of your organisation
+          </h1>
+        </legend>
+
+        <div class="govuk-radios">
+
+          <div class="govuk-radios__item">
+            <%= f.radio_button :mission, "black_or_minority_ethnic_led", class: "govuk-radios__input" %>
+            <%= f.label :organisation_mission_black_or_minority_ethnic_led, "Black or minority ethnic led", class: "govuk-label govuk-radios__label" %>
```
gidsg: The labels don't match the inputs because Rails adds an extra `organisation` namespace to demote the model being updated. I think we could add as partial which sets them correctly?

gidsg:
Seems like we are not getting a lot of value from the `wicked` stuff in this implementation, as discussed. So I think we should try moving these together in one controller so we can get more re-use. Or removing `wicked` and doing the routing by hand?

Gemfile
stuartmccoll: Removed library, so this is now removed from the `Gemfile` entirely.

app/controllers/organisation/organisation_numbers_controller.rb
stuartmccoll: Done 👍.

app/controllers/organisation/organisation_numbers_controller.rb
stuartmccoll: Wicked library removed, so this no longer exists in the code 👍.

app/controllers/organisation/organisation_mission_controller.rb
stuartmccoll: This is gone now 👍.

Gemfile.lock
```diff
@@ -267,6 +267,8 @@ GEM
     websocket-driver (0.7.1)
       websocket-extensions (>= 0.1.0)
     websocket-extensions (0.1.4)
+    wicked (1.3.4)
```
gidsg: this is still in here. I think if we run `bundle install` it should remove it.

stuartmccoll: Done :).

app/controllers/home_controller.rb
```diff
@@ -1,7 +1,42 @@
...
+  # Before rendering the page, we check whether the user belongs to an
+  # organisation and create a new Organisation instance if not. This is
+  # then used to provide the organisation UUID in the URL.
   def show
+
+    if current_user.organisation
+
+      logger.info 'Existing organisation ' + current_user.organisation.id +
+                      ' found for ' + current_user.id.to_s
+      @start_now_link_value = organisation_organisation_summary_get_path(current_user.organisation.id)
```
gidsg: we can remove the `.id` here and Rails should still resolve the path correctly.

app/controllers/home_controller.rb
```diff
@@ -1,7 +1,42 @@
...
+    else
+
+      logger.info 'No existing organisation found for user ID: ' + current_user.id.to_s
+
+      logger.debug 'Creating organisation for user ID: ' + current_user.id.to_s
+
+      @organisation = Organisation.create
+
+      logger.debug 'Finished creating organisation ' + @organisation.id +
+                       ' for user ID: ' + current_user.id.to_s
```
gidsg: Very minor, we can use string interpolation to make this a bit more readable e.g.
```
"Finished creating organisation #{@organisation.id} for user ID: #{current_user.id}"
```

app/controllers/organisation/organisation_about_controller.rb
```diff
@@ -1,4 +1,30 @@
 class Organisation::OrganisationAboutController < ApplicationController
-  def organisation_about
+  include OrganisationHelper
+  before_action :authenticate_user!, :set_organisation
+
+  def show
+    render :'organisation/organisation_about/about'
```
gidsg: could we use a Rails url helper to derive the route? This would make it more robust in the event of the URL changing.
e.g. `organisation_organisation_about_get_path`?

app/controllers/organisation/organisation_signatories_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    redirect_to :organisation_organisation_summary_get
+
+  end
+
+  private
+
+  def set_legal_signatories
+
+    @legal_signatories = LegalSignatory.where(organisation_id: @organisation.id)
```
gidsg: I think we could simplify this code by doing the query through association e.g. 
`@organisation.legal_signatory` will resolve to the associated entity.

app/views/organisation/organisation_signatories/signatories.html.erb
```diff
@@ -0,0 +1,36 @@
...
+        <h2 class="govuk-heading-m" aria-label="Heading">Legal signatory <%= i+1 %></h2>
+
+        <div class="govuk-form-group">
+          <%= signatory_field.label :name, "Full name", class: "govuk-label" %>
+          <%= signatory_field.text_field :name, class: "govuk-input govuk-input--width-20" %>
+        </div>
+
+        <div class="govuk-form-group">
+          <%= signatory_field.label :email_address, "Email address", class: "govuk-label" %>
+          <%= signatory_field.text_field :email_address, class: "govuk-input govuk-input--width-20" %>
```
gidsg: we could use `email_field` for this. https://api.rubyonrails.org/v6.0.1/classes/ActionView/Helpers/FormHelper.html#method-i-email_field

gidsg:
@stuartmccoll Looks good to me if you are able to resolve the merge conflict.

PR #28, opened 2019-12-16
Send Devise emails via Notify

gidsg:
This is part of https://trello.com/c/JM1iwUIo/915-large-feature-email-notifications-via-govuk-notify I will do the other parts in separate future PRs.


config/environments/production.rb
```diff
@@ -110,4 +110,11 @@
   # config.active_record.database_selector = { delay: 2.seconds }
   # config.active_record.database_resolver = ActiveRecord::Middleware::DatabaseSelector::Resolver
   # config.active_record.database_resolver_context = ActiveRecord::Middleware::DatabaseSelector::Resolver::Session
+  # Send emails via notify
+  # TODO: replace with production host
+  config.action_mailer.default_url_options = { :host => 'http://example.com' }
```
stuartmccoll: Would `{host: 'http://example.com'}` work here instead of the `=>` notation?

PR #29, opened 2019-12-16
Make users confirm email

gidsg:
https://trello.com/c/KkmAxQK2/871-implement-a-confirm-email-step-in-the-account-registration-flow
Note: locally you should see the confirm URL in the console you can use to test this.
Also, this uses the default flow provided by Devise which is quite clunky and seems to make you re-login after confirming. We should review this flow with @paulmsmith to see if it can be improved.


gidsg:
@stuartmccoll I think they should be unrelated as this only adds things to the user model.

PR #31, opened 2019-12-16
Feature/nlhf branding

paulmsmith:


app/views/partials/_header.html.erb
```diff
@@ -0,0 +1,10 @@
+<header class="nlhf-header-global">
+    <div class="govuk-width-container">
+        <div class="nlhf-header-global__logo">
+            <img src="<%= image_path("logo_on_coloured-background.svg") %>" alt="National Lottery Heritage Fund">
```
gidsg: we could use `image_tag` helper here: https://api.rubyonrails.org/classes/ActionView/Helpers/AssetTagHelper.html#method-i-image_tag
e.g
```
image_tag("logo_on_coloured-background.svg", alt: "National Lottery Heritage Fund") 
``` 
should be equivalent I believe.

paulmsmith: @gidsg - In this instance you're right but I might want total control of the image markup in future and on other image tags/elements used so I wonder you think we should we stick to `image_path` so not to have a mixture of both?

paulmsmith:
Shall I merge @stuartmccoll?

PR #32, opened 2019-12-17
Feature/non cash contributions

PushBackNeil:


config/routes.rb
```diff
@@ -1,6 +1,9 @@
 Rails.application.routes.draw do
 
 
+  namespace :project do
```
stuartmccoll: Not sure what's happened here, but we've ended up with two `:project` namespaces.

Looks like this already exists in the other `:project` namespace, albeit the method being called is one character different.

app/assets/stylesheets/application.scss
```diff
@@ -23,10 +23,5 @@
...
 
-// ############################################################################
-// NLHF SCSS Imports (namespace 'nlhf-')
-// ############################################################################
-@import "helpers/all";
-@import "tools/all";
-@import "components/all";
-@import "utilities/all";
+// other NLHF imports below here
+@import "nlhf";
```
stuartmccoll: Has something gone wrong with merging `master` in here? Looking at the latest copy of `master`, this seems to be taking @paulmsmith's changes back out.

app/assets/stylesheets/_nlhf.scss
```diff
@@ -0,0 +1,290 @@
+
```
stuartmccoll: Looks like this is reverting @paulmsmith's changes from `master` too.

config/routes.rb
PushBackNeil: @stuartmccoll good spot, rails g strikes again!

app/assets/stylesheets/_nlhf.scss
PushBackNeil: Looks like the merge from master went wrong, I will fix it.

PushBackNeil: @paulmsmith its strange the files Paul added were there.

PR #36, opened 2019-12-17
Validation checks for organisation type form

stuartmccoll:
Raised issue #40 to revisit this as a piece of technical debt at a later date.

### DIFF
--- a/app/controllers/organisation/organisation_type_controller.rb
+++ b/app/controllers/organisation/organisation_type_controller.rb
@@ -9,6 +9,7 @@ class Organisation::OrganisationTypeController < ApplicationController
   end
 
   def update
+    @organisation.validate_org_type
 
     if !params[:organisation].present?
 
--- a/app/controllers/organisation/organisation_type_controller.rb
+++ b/app/controllers/organisation/organisation_type_controller.rb
@@ -9,7 +9,7 @@ class Organisation::OrganisationTypeController < ApplicationController
   end
 
   def update
-    @organisation.validate_org_type
+    @organisation.validate_org_type = true
 
     if !params[:organisation].present?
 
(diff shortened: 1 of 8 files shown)


app/views/layouts/application.html.erb
```diff
@@ -22,7 +22,6 @@
           <%= yield %>
         </div>
       </div>
-    </div>
```
gidsg: is this change intentional?

app/controllers/organisation/organisation_type_controller.rb
```diff
@@ -2,49 +2,47 @@ class Organisation::OrganisationTypeController < ApplicationController
...
 
-        flash[:alert] = @@error_message
+    else
 
-      else
+      logger.debug "Organisation type not found when attempting to update organisation ID: " +
+                       "#{@organisation.id}"
 
-        logger.debug "Finished updating organisation ID: #{@organisation.id}"
+      flash[:alert] = @organisation.errors[:org_type][0]
```
gidsg: I think we could make an error handler that iterates through this hash, displaying the errors at the correct location, which we could re-use as a partial, rather than writing bespoke code for each page. But happy to do this for now and we can re-visit potentially.

app/views/layouts/application.html.erb
stuartmccoll: I'm not sure how this branch ended up with an extra closing `div` tag, but this line is taking it out - looking at the file in full, the number of opening and closing tags matches what's in `master` now.

app/views/organisation/organisation_type/type.html.erb
```diff
@@ -1,8 +1,26 @@
...
+      </ul>
+    </div>
+  </div>
+
+<% end %>
 
 <%= form_for @organisation, url: organisation_organisation_type_put_url, method: :put do |f|  %>
 
-  <div class="govuk-form-group">
+  <div class="govuk-form-group <% if flash[:alert] %>govuk-form-group--error<% end %>">
```
gidsg: We could express this slightly more concisely as:
`<div class="govuk-form-group <% "govuk-form-group--error"  if flash[:alert] %>">`

PR #44, opened 2019-12-19
Added validation to organisation mission form

stuartmccoll:


app/models/organisation.rb
```diff
@@ -17,8 +23,24 @@ def validate_charity_number?
...
-  def validate_org_type?
-    validate_org_type == true
+  def validate_mission?
+    validate_mission == true
+  end
+
+  # Custom validator to determine whether any of the items in the incoming mission array
+  # are not included in the expected list of options
+  def validate_mission_array
+    if !mission.nil?
```
gidsg: we could use `mission.present?`

PR #45, opened 2019-12-19
Feature/toolbar

paulmsmith:
This PR adds a simple "toolbar' at the top of the page for users to be able to sign out or jump to their account page. It is static so will need wiring up. You could wire up in this branch or merge and then wire it up in a new branch.


app/views/layouts/application.html.erb
```diff
@@ -3,7 +3,7 @@
 <head>
   <title><%= yield(:page_title) %> - National Lottery Heritage Fund</title>
   <meta charset="utf-8" />
-  <title>FundingFrontendRuby</title>
```
paulmsmith: This was a duplicate so removed it.

app/views/layouts/application.html.erb
```diff
@@ -3,7 +3,7 @@
...
+  <meta name="viewport" content="width=device-width, initial-scale=1, viewport-fit=cover">
```
paulmsmith: Helps mobile users have the correct viewport size/scale

app/views/partials/_header.html.erb
```diff
@@ -5,6 +5,3 @@
         </div>
     </div>
 </header>
-<div class="govuk-width-container">
-    <a href="<%=request.referrer %>" class="govuk-back-link" aria-label="go back">Back</a>
-</div>
```
paulmsmith: Removed for now. This might come back but somewhere else in the page design

PR #46, opened 2019-12-19
Feature/project cost example data

PushBackNeil:


app/views/project/project_costs/project_costs.html.erb
```diff
@@ -64,21 +65,91 @@
...
-              <span class="nlhf-summary__total__title-val">£0</span>
-            </h3>
-          </div>
+          <% if @total_costs == 0 %>
+            <h3 class="govuk-heading-m">You have not added any costs yet</h3>
+            <p class="govuk-body">Each cost will be shown here with a running total</p>
+
+          <% else %>
+            <table class="govuk-table">
+              <!--<caption class="govuk-table__caption">Your project costs</caption>-->
```
stuartmccoll: Do we need this? Wasn't sure if it was meant to be uncommented or if we're safe to remove it from the code completely rather than leave it in and commented out.

PushBackNeil: I will take it out @stuartmccoll

PR #49, opened 2019-12-20
Added validation for organisation legal signatories form

stuartmccoll:
This pull request adds validation to the legal signatories form for an organisation.

I haven't split the summary display of errors out into a partial, as this might be the only place that we need to loop through an array of objects before looping through their error messages. If this is something that we end up having to do again elsewhere, then this piece of code can be refactored into a partial which takes a set of parameters.

### DIFF
--- a/app/views/organisation/organisation_signatories/signatories.html.erb
+++ b/app/views/organisation/organisation_signatories/signatories.html.erb
@@ -15,7 +15,7 @@
         <% @legal_signatories.each do |signatory| %>
           <% signatory.errors.each do |attr, msg| %>
             <li>
-              <a data-turbolinks="false" href="#legal_signatories_<%= signatory.id %>_<%= attr %>"><%= msg %></a>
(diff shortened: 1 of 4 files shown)


app/views/organisation/organisation_signatories/signatories.html.erb
```diff
@@ -1,36 +1,71 @@
...
+      <ul class="govuk-list govuk-error-summary__list">
 
-        <div class="govuk-form-group">
-          <%= signatory_field.label :name, "Full name", class: "govuk-label" %>
-          <%= signatory_field.text_field :name, class: "govuk-input govuk-input--width-20" %>
-        </div>
+        <% @legal_signatories.each do |signatory| %>
+          <% signatory.errors.each do |attr, msg| %>
+            <li>
+              <a data-turbolinks="false" href="#legal_signatories_<%= signatory.id %>_<%= attr %>"><%= msg %></a>
```
gidsg: could we use string interpolation to make this value easier to read?

stuartmccoll: Fixed in latest commit 👍.

PR #52, opened 2019-12-30
add routing for account links

gidsg:
Replace dummy account links with devise routes.


app/views/partials/_toolbar.html.erb
```diff
@@ -4,10 +4,14 @@
             <div class="govuk-grid-column-full">
                 <ul class="nlhf-toolbar__items">
                     <li class="nlhf-toolbar__item">
-                        <a href="#account" class="nlhf-toolbar__link">Your account</a>
+                      <%= link_to('Your account', edit_user_registration_path, class: 'nlhf-toolbar__link') %>
```
stuartmccoll: Should we only display this if a user is already signed in? More of a content/design question I suppose, happy to merge and pick it up with @paulmsmith when he's back from leave.

gidsg: OK I added that condition in for now. @paulmsmith can review when he's back.

app/views/partials/_toolbar.html.erb
```diff
@@ -3,11 +3,17 @@
...
                 <ul class="nlhf-toolbar__items">
+                  <% if user_signed_in? %>
                     <li class="nlhf-toolbar__item">
-                        <a href="#account" class="nlhf-toolbar__link">Your account</a>
+                      <%= link_to('Your account', edit_user_registration_path, class: 'nlhf-toolbar__link') %>
                     </li>
+                    <% end %>
                     <li class="nlhf-toolbar__item">
-                        <a href="#account" class="nlhf-toolbar__link">Sign out</a>
+                      <% if user_signed_in? %>
```
stuartmccoll: I think you only need one `if/else` now - `if user_signed_in?` then show the 'Your account' and 'Logout' links, `else` show the 'Login' link.

gidsg: I'm not sure if it makes sense to show the login link at all, as I think the user will need to be signed in on all pages except the sign-in page. May make more sense to hide the whole toolbar if the user is not signed in?

stuartmccoll: Yeah, you're right, all of the organisation pages are sat behind the auth gate. Agreed that the toolbar is probably superfluous if you're not signed in, as you'll always be sat on a page with either 'sign up' or 'login' links.

Let's merge what's there and then see what @paulmsmith thinks about hiding the toolbar completely 🙂.

PR #53, opened 2019-12-30
First iteration of postcode lookup in organisation flow

stuartmccoll:


app/controllers/organisation/organisation_about_controller.rb
```diff
@@ -1,30 +1,159 @@
...
+      render :about
+
+    end
 
   end
 
   private
 
+  def set_api_key
+    IdealPostcodes.api_key = ENV['IDEAL_POSTCODES_API_KEY']
```
gidsg: Can we move this to configuration e.g. https://edgeguides.rubyonrails.org/configuring.html#custom-configuration and use `env.fetch('IDEAL_POSTCODES_API_KEY')`?
that way we can check this at startup rather than a runtime error if it is not set.

app/controllers/organisation/organisation_about_controller.rb
```diff
@@ -1,30 +1,159 @@
...
+  end
+
+  # Takes a given address in the params hash and looks this up within
+  # the IdealPostcodes API
+  def lookup_address
+
+    # TODO: Add error handling around API exceptions
+
+    begin
+      @address_response = IdealPostcodes::Address.lookup params[:address]
```
gidsg: Currently, we are doing 2 lookups in order to display the results on the subsequent page (which has a cost implication as we pay per-lookup.) Could we add a TODO to cache the first query and use that to lookup on the following page if available?

app/controllers/organisation/organisation_about_controller.rb
```diff
@@ -1,30 +1,159 @@
...
   end
 
+  # Makes a request to the IdealPostcodes API to retrieve a list of addresses
+  # matching the given postcode.
+  # If no postcode or an invalid postcode is found, this method redirects the user
+  # back to the :postcode_lookup view.
+  def display_address_search_results
+
+    if !params['postcode']['lookup'].present?
+
```
gidsg: Can we add a TODO to add a format validator? e.g. we can use a library like: https://github.com/globaldev/going_postal

app/controllers/organisation/organisation_about_controller.rb
```diff
@@ -1,30 +1,159 @@
...
     render :about
   end
 
+  # Makes a request to the IdealPostcodes API to retrieve a list of addresses
+  # matching the given postcode.
+  # If no postcode or an invalid postcode is found, this method redirects the user
+  # back to the :postcode_lookup view.
+  def display_address_search_results
+
+    if !params['postcode']['lookup'].present?
```
gidsg:
```suggestion
    unless params['postcode']['lookup'].present?
```
we can use this syntax instead.

app/views/partials/_summary_errors_flash.html.erb
```diff
@@ -0,0 +1,21 @@
...
+    There is a problem
+  </h2>
+
+  <div class="govuk-error-summary__body">
+
+    <ul class="govuk-list govuk-error-summary__list">
+
+      <% flash[:errors].each do |attr, msg| %>
+        <li>
+          <a data-turbolinks="false" href="#<%= first_form_element %>"><%= msg %></a>
```
gidsg: Seems like this would always link to the first element with error? Is that correct?

stuartmccoll: It is at the minute, but only because this is currently the _only_ place we're checking `flash[:errors]`, and we know that we only set one error message. I was supposed to stick a comment in here saying we could make it more reusable if we end up needing to use it elsewhere!

config/environments/production.rb
```diff
@@ -112,8 +112,14 @@
   # config.active_record.database_resolver_context = ActiveRecord::Middleware::DatabaseSelector::Resolver::Session
   # Send emails via notify
   config.action_mailer.default_url_options = { host: "https://#{JSON.parse(ENV['VCAP_APPLICATION'])['application_uris'][0]}" }
-    config.action_mailer.delivery_method = :notify
-    config.action_mailer.notify_settings = {
-        api_key: ENV['NOTIFY_API_KEY']
-    }
+  config.action_mailer.delivery_method = :notify
+  config.action_mailer.notify_settings = {
```
gidsg: Rails convention for these is to put these directly under `config` or under `config.x` if nested as per: https://edgeguides.rubyonrails.org/configuring.html#custom-configuration
I think this is to prevent conflicts with library generated keys
so e.g. either
`config.notify_api_key`
or 
`config.x.notify.api_key`
although in these cases I don't think we need the nesting as each item only has one value.

config/environments/production.rb
```diff
@@ -114,10 +114,10 @@
   config.action_mailer.default_url_options = { host: "https://#{JSON.parse(ENV['VCAP_APPLICATION'])['application_uris'][0]}" }
   config.action_mailer.delivery_method = :notify
   config.action_mailer.notify_settings = {
-      api_key: ENV['NOTIFY_API_KEY']
+      api_key: ENV.fetch("NOTIFY_API_KEY")
   }
 
-  config.ideal_postcodes = {
+  config.x.ideal_postcodes = {
       settings: {
```
gidsg: I don't think we should include this settings hash, that is only included above because the library expects it. So, just assign to 
`config.x.ideal_postcodes.api_key`

PR #55, opened 2019-12-31
Added functionality to project title page and some temporary model creation

stuartmccoll:


app/controllers/project/new_project_controller.rb
```diff
@@ -1,4 +1,16 @@
 class Project::NewProjectController < ApplicationController
+  before_action :authenticate_user!
+
   def new_project
   end
+
+  # TODO: This is a temporary method to create a new project in order to
+  #       pass this to later routes. Once this has been wired up properly,
+  #       this method can be removed.
+  def temp_create_new_project
```
stuartmccoll: Note that this is a temporary method to create the project object. This will be stripped out and replaced with the correct functionality once we have designs in place for the project dashboard and creation flow.

app/views/project/new_project/new_project.html.erb
```diff
@@ -1,2 +1,8 @@
 <h1>Project::NewProject#new_project</h1>
 <p>Find me in app/views/3-10k/project/new_project/new_project.html.erb</p>
+
+<%= form_tag(three_to_ten_k_project_temp_create_new_url, {method: :put}) do %>
```
stuartmccoll: Note that this is temporary routing for now, we'll revisit this page when we have designs in place.

PR #56, opened 2019-12-31
Add cash contributions page

gidsg:
Add cash contributions using a separate model (as they require unique IDs for Salesforce payload)
Populates the model associated with the project using Rails [nested form](https://guides.rubyonrails.org/form_helpers.html#nested-forms) functionality.
Uses the same basic server-side file upload mechanism as the supporting evidence page.


db/migrate/20191230135515_create_cash_contributions.rb
```diff
@@ -0,0 +1,10 @@
+class CreateCashContributions < ActiveRecord::Migration[6.0]
+  def change
+    create_table :cash_contributions, id: :uuid do |t|
+      t.integer "amount"
```
stuartmccoll: Are we always expecting a whole number to be entered against this? If not, should we be using `float` or `decimal` instead?

app/controllers/project/project_cash_contribution_controller.rb
```diff
@@ -1,29 +1,20 @@
...
-    end
+  def project_cash_contribution
+    @cash_contribution = @project.cash_contributions.build
   end
 
-  def add_cash_contribution
-    # TODO save cash_contribution
+
+  def put
+    @project.update(project_params)
```
stuartmccoll: Is it worth leaving a `TODO` in here to pick up the validation work?

app/views/project/project_cash_contribution/project_cash_contribution.html.erb
```diff
@@ -32,50 +30,42 @@
...
               </div>
             </div>
             <div class="nlhf-add-item__row">
               <div class="govuk-form-group">
-                <label class="govuk-label" for="cash-cont-amount">
-                  Amount
-                </label>
-                <input class="govuk-input" id="cash-cont-amount" name="cash-cont-amount" type="text" autocomplete="off">
+              <%= cc.label :amount, class: 'govuk-label' %>
+              <%= cc.text_field :amount, class: 'govuk-input' %>
```
stuartmccoll: We could make this a `number_field` instead, as we're tying it to some kind of number-based column type.

db/migrate/20191230135515_create_cash_contributions.rb
gidsg: Yes, I believe we always expect whole number in Pounds but we should indicate that in the UI somehow. One for @paulmsmith to think about.

PR #57, opened 2020-01-01
Implementation of Project Dates functionality

stuartmccoll:


app/controllers/project/project_dates_controller.rb
```diff
@@ -1,4 +1,93 @@
...
+                                    :start_date_month,
+                                    :start_date_year,
+                                    :end_date_day,
+                                    :end_date_month,
+                                    :end_date_year)
+  end
+
+  # This method checks to determine whether or not the project length
+  # is greater than one year
+  def is_project_length_greater_than_one_year(first_date, second_date)
```
gidsg: Could we implement this as a custom validator? i.e. in the style described in https://guides.rubyonrails.org/active_record_validations.html#custom-validators

gidsg:
I think maybe because of the way we are discarding the flash, if there are errors, the user input is lost.

gidsg:
Would be nice if a standard validation message is shown, to display the attribute name for context
Screenshot 2020-01-02 at 11 29 41
As this is unrelated to this specific PR I can raise an Issue for it.

gidsg:
I think there was talk of enforcing that the start date is at least 8 weeks from now, as they cannot get a decision before then. However, not sure what was decided on this, as the current form does not do this I would say it is not critical but @Ellayanor can confirm.

app/controllers/project/project_dates_controller.rb
```diff
@@ -1,4 +1,93 @@
...
+
+    logger.debug "Calculating length of project in days for project ID: #{@project.id}"
+
+    length_in_days = (second_date - first_date).to_i
+
+    logger.debug "Project length in days of #{length_in_days} for project ID: #{@project.id}"
+
+    if length_in_days > 365
```
gidsg: Could we express this more simply as `return length_in_days > 365`?

gidsg:
very minor but the the anchor links are broken on these validation errors
Screenshot 2020-01-02 at 12 00 41
I think we could resolve this by adding `project_start_date` and `project_end_date` `id` attributes to the relevent `legend` elements or in a wrapper `div`.

gidsg:
> Could we implement this as a custom validator? i.e. in the style described in

I think this commit https://github.com/heritagefund/funding-frontend/pull/57/commits/117d13baf8d4c3db8fcd5a51c16237cf22b776f2 broke the creation of a new project as it tries to apply the validator on that page. I think the simplest may be to revert this commit and go back to the old style specific to the controller.

app/controllers/project/project_dates_controller.rb
```diff
@@ -1,4 +1,99 @@
...
+
+  end
+
+  private
+
+  def store_values_in_flash
+
+    params[:project].each do | key, value |
+      flash[key] = value.empty? ? "a" : value
+    end
```
gidsg: We could express this as:
```
    params[:project].reject{|_k,v| v.empty?}.each do | key, value |
      flash[key] =  value
    end
```
I don't _think_ we need to set the literal `"a"` value as the template handles the `nil` value fine but let me know if I've missed something.

stuartmccoll: Oops, didn't mean to leave the value as `"a"`. I've changed it to `0` - this ensures that if the project already has start and end dates assigned and a user then submits the form with one or more of the inputs empty, then the error is flagged and the input field is then filled with a `0`. Without this, I was ending up with the value from the model being pre-filled in the input, because I couldn't get the `flash` to play nicely with an empty value for a key.

There's possibly a better way to do it, but this gives us a working starting point.

stuartmccoll:
> > Could we implement this as a custom validator? i.e. in the style described in
> 
> I think this commit [117d13b](https://github.com/heritagefund/funding-frontend/commit/117d13baf8d4c3db8fcd5a51c16237cf22b776f2) broke the creation of a new project (from `/3-10k/project/new`) as it tries to apply the validator on that page. I think the simplest may be to revert this commit and go back to the old style specific to the controller.

Fixed 👍.

app/controllers/project/project_dates_controller.rb
gidsg: OK, maybe add a `TODO` here and we'll take another look at this if we get a chance?

PR #58, opened 2020-01-02
Added functionality to project description page

stuartmccoll:
Note that I've left a `TODO` on wiring up the JavaScript word count. I've tried dropping in the sample component from the GOV.UK Design System and this didn't work, which makes me wonder if we're enabling JavaScript in the application properly - I think it's worth looking at this separately, perhaps when @paulmsmith and @PushBackNeil are back from leave.


app/views/project/project_description/show.html.erb
```diff
@@ -0,0 +1,44 @@
...
+
+      <%= render partial: "partials/form_input_errors",
+                 locals: {form_object: @project,
+                          input_field_id: :description} if @project.errors[:description].any? %>
+
+      <%= f.text_area :description, rows: 10,
+                      class: "govuk-textarea govuk-js-character-count #{'govuk-input--error' if @project.errors[:description].any?}",
+                      "aria-describedby" => "project_description-info project_description-hint" %>
+
+      <%# TODO: Fix word count JS implementation %>
```
gidsg: Yes, I had the same issue in https://github.com/heritagefund/funding-frontend/blob/a03f6b6055636b57ccacbcaf94af77c003de99d8/app/views/project/project_support_evidence/project_support_evidence.html.erb#L25

stuartmccoll: Referenced in a separate issue to make doubly sure that we don't forget 👍.

PR #65, opened 2020-01-03
Added functionality to project difference page

stuartmccoll:
This pull request adds functionality to the 'What difference will your project make page?'.

As well as this, I have refactored the calculation of word count into a custom validation method which can be called for any attribute by passing in the relevant values.


app/models/project.rb
```diff
@@ -33,7 +34,22 @@ class Project < ApplicationRecord
 
     validates :project_title, presence: true, length: { maximum: 255 }, if: :validate_title?
     validates :description, presence: true, if: :validate_description?
-    validate :validate_description_length, if: :validate_description?
+
+    validate do ||
```
gidsg:
```suggestion
    validate do
```
I don't _think_ these pipes are doing anything.

app/models/project.rb
```diff
@@ -33,7 +34,22 @@ class Project < ApplicationRecord
...
+        validate_length(
+            :description,
+            500,
+            "Project description must be 500 words or fewer"
+        ) if validate_description?
+    end
+
+    validate do ||
```
gidsg:
```suggestion
    validate do
```
ibid

Issue #62, opened 2020-01-02
display flash[:notice] and flash[:alert] messages on Devise pages


gidsg:
Currently we are not displaying notifications on the sign-in/sign-up pages. We need to include `flash[:notice]` and `flash[:alert]` in these templates so these are visible.
CC @paulmsmith

gidsg:
I think this is causing this issue on a subsequent page, where the notice is being retained and displayed erroneously here. CC @stuartmccoll 
<img width="584" alt="Screenshot 2020-01-03 at 10 45 29" src="https://user-images.githubusercontent.com/1764158/71719457-521cd180-2e16-11ea-9717-8d83550d0748.png">

stuartmccoll:
I've raised #68 to fix the sign-in message being displayed.

paulmsmith:
Is this something @PushBackNeil can look at today once my branch has been merged in?

gidsg:
I think @PushBackNeil could pick this up. We should use one of the approaches here to restrict it to Devise so it does not affect other pages: https://github.com/plataformatec/devise/wiki/How-To:-Create-custom-layouts

PR #68, opened 2020-01-03
Removed usage of flash.notice on project dates page

stuartmccoll:
> I think this is causing this issue on a subsequent page, where the notice is being retained and displayed erroneously here. CC @stuartmccoll
> Screenshot 2020-01-03 at 10 45 29

This fixes the above issue referenced in #62.

Note that we'll have to avoid using the built-in `flash` keys as these are used within the registration/login flow.


gidsg:
> Note that we'll have to avoid using the built-in flash keys as these are used within the registration/login flow.

I think normally it would be fine to use those keys but because of #62 the flash is not getting cleared once devise raises a message.

PR #71, opened 2020-01-03
Added address lookup and save functionality to project location

stuartmccoll:
This pull request adds functionality to lookup an address by postcode and to save these details as the project address. Note that this is duplicated code from the same functionality within the organisation journey. I'll raise a separate issue to revisit this functionality and refactor it into a single place - #72


app/controllers/project/project_location_controller.rb
```diff
@@ -1,10 +1,121 @@
+require 'ideal_postcodes'
+require 'pry'
```
gidsg: we should remove this as it is just used for debugging.
```suggestion

```

PR #75, opened 2020-01-03
Added functionality to project heritage page

stuartmccoll:
This pull request adds functionality to the 'The heritage of your project: How do you plan to make it available once the project is over?' page.


app/views/project/project_heritage/show.html.erb
```diff
@@ -0,0 +1,52 @@
+<% content_for :page_title,
```
gidsg: We use this pattern a bunch of times for the page title, maybe we can pull it out into a helper? Doesn't have to be in this PR though.

scaff.sh
```diff
@@ -13,7 +13,7 @@ rails g controller project/project_description project_description
 rails g controller project/project_permission project_permission
 rails g controller project/project_difference project_difference
 rails g controller project/project_matters project_matters
-rails g controller project/project_availability project_availability
+rails g controller project/project_heritage project_heritage
```
gidsg: I think we should delete this file as it is just an artifact of the dev process. Doesn't have to be in this PR though.

app/views/project/project_heritage/show.html.erb
stuartmccoll: Agreed - created a new issue for it so we don't forget.

scaff.sh
stuartmccoll: Will take it out in the next pull request 👍.

PR #86, opened 2020-01-07
Wired up project dashboard functionality

stuartmccoll:
This commit changes the flow of the service journey to match the current service journey flow diagram.

Upon logging in, a user will be directed to the projects dashboard page. From here, clicking through to the Apply for Funding page will initially direct them to create an organisation if they have not alreaady done so. The end of the organisation creation journey will direct them immediately to the create a project journey. For users that already have an organisation, they will be directed straight to the create a project journey.


paulmsmith:
Tested this locally and it works a treat for me. Unstyled yet but I will merge this and do that work in separate branch.

app/controllers/dashboard_controller.rb
```diff
@@ -2,7 +2,9 @@ class DashboardController < ApplicationController
   before_action :authenticate_user!
 
   def show
-    @user = current_user
+
+    @projects = Project.all.where(user_id: current_user.id)
```
gidsg:
```suggestion
    @projects = current_user.projects
```
I think we can express this more concisely through association.

app/views/dashboard/show.html.erb
```diff
@@ -1,4 +1,34 @@
...
+
+<% else %>
+
+  <ul class="govuk-list">
+
+    <% @projects.each do| project | %>
+
+    <li>
+      <a href="<%= three_to_ten_k_project_title_get_url(project_id: project.id) %>" class="govuk-link govuk-link--no-visited-state">
+        <%= project.project_title.present? ? project.project_title : "Untitled project" %>
```
gidsg:
```suggestion
        <%=  project.project_title || "Untitled project" %>
```
we could express this more concisely

test/controllers/dashboard_controller_test.rb
```diff
@@ -8,9 +8,4 @@ class DashboardControllerTest < ActionDispatch::IntegrationTest
     post user_session_url
   end
 
-  test "should get show" do
-    get dashboard_show_url
-    assert_response :success
-  end
-
 end
```
gidsg: this test file is now just setup with no tests. Maybe we should delete the whole file?

PR #88, opened 2020-01-07
Account creation pages

paulmsmith:
Styles the login, reset-password, re-send confirmation, sign-up pages.


app/views/devise/shared/_error_messages.html.erb
```diff
@@ -1,15 +1,14 @@
 <% if resource.errors.any? %>
-  <div id="error_explanation">
-    <h2>
-      <%= I18n.t("errors.messages.not_saved",
```
gidsg: Maybe best instead to update the localised string which is defined in: https://github.com/heritagefund/funding-frontend/blob/master/config/locales/devise.en.yml#L63
this is because we will then be able to use this localised string for Welsh translation later on.

paulmsmith: What confused me here was the `errors.messages.not_saved` because this is a partial used elsewhere it didn't seem to make sense in other places.

gidsg: We could just update the value to _There was a problem_ which is what you are using?

paulmsmith: @gidsg Sure for now I can do that.

PR #91, opened 2020-01-07
Display account error and info messages.

### LINKED ISSUE #62: display flash[:notice] and flash[:alert] messages on Devise pages

gidsg:
This displays errors and notifications and errors from devise.
Fixes #62 
@paulmsmith we need to handle the information messages generated by devise. For now I added them in the error area unstyled but they may be more suited to an overlay style or something which denotes success.. 
Please can you figure out a design for this.
Screenshot 2020-01-07 at 13 16 03


gidsg:
As discussed, @stuartmccoll can you review and we will merge this as-is for now to support user research.

PR #93, opened 2020-01-07
Added functionality to project permission page

stuartmccoll:
This commit adds functionality to the 'Do you need permission from anybody else to start your project?' page. Note that at the moment this _only_ includes server-side functionality. We should revisit this at a later date to add client-side functionality to display hidden fields if a user has JavaScript enabled, etc.


app/controllers/project/project_permission_controller.rb
```diff
@@ -1,4 +1,141 @@
...
+      @project.errors.add(
+          :permission_type,
+          "Select an option to tell us whether you need permission from others to do your project"
+      )
+
+      render :show
+
+    else
+
+      if flash.empty?
```
gidsg: Are we using this as a proxy for no errors? If so can we use a more explicit way of checking for this e.g.
`unless project.errors.any?`
let me know if I've missed something?

app/views/project/project_permission/show.html.erb
```diff
@@ -0,0 +1,95 @@
...
+        </div>
+
+        <div class="govuk-radios__item">
+          <%= f.radio_button :permission_type, "not_sure", class: "govuk-radios__input" %>
+          <%= f.label :permission_type_not_sure, "Not sure", class: "govuk-label govuk-radios__label" %>
+        </div>
+
+        <% if flash[:show_permission_description_for_type_not_sure] %>
+
+          <%= render partial: "partials/project_permission_description",
```
gidsg: I think we could express these kind of custom fields more consisely using custom form builder elements i.e. https://api.rubyonrails.org/v6.0/classes/ActionView/Helpers/FormBuilder.html but happy to raise an issue and look at this post launch

stuartmccoll: Makes sense to me, I've raised a new issue for it 👍.

app/controllers/project/project_permission_controller.rb
stuartmccoll: The `flash` use in this controller is basically driving whether or not the page needs to re-render with the `textarea` displayed for one of the relevant radio buttons. Using `flash` as there aren't actually any errors at such at this point, but we need something to indicate whether or not the relevant `textarea` should be displayed.

Once we tackle the client-side display of hidden fields, etc, this code will only be relevant to the non-JavaScript journey and will need some refactoring to _only_ trigger if JavaScript is disabled.

stuartmccoll:
Closing this, as I've raised a new pull request based on `master` which encompasses changes made to the stubbed project permissions page.

PR #96, opened 2020-01-08
Updated to_salesforce_json method of project model

stuartmccoll:
This pull request adds some of the missing model fields to the `to_salesforce_json` method for the `project` model, with dummy data and real data where this is available.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -166,6 +166,67 @@ class Project < ApplicationRecord
                 json.set!('username', self.user.email)
             end
             json.application do
+                json.set!('projectName', self.project_title)
+                json.projectDateRange do
+                    json.startDate self.start_date
+                    json.endDate self.end_date
+                end
+                json.projectAddress do
+                    json.line1 self.line1
+                    json.line2 self.line2
+                    json.line3 self.line3
+                    json.county self.county
+                    json.townCity self.townCity
+                    json.projectPostcode self.postcode
+                end
+                json.set!('yourIdeaProject', self.description)
+                json.set!('projectDifference', self.difference)
+                json.set!('projectCommunity', self.matter)
+                json.set!('projectOrgBestPlace', self.best_placed_description)
+                json.set!('projectAvailable', self.heritage_description)
+                json.set!('projectOutcome1', 'Dummy data')
+                json.set!('projectOutcome2', 'Dummy data')
+                json.set!('projectOutcome3', 'Dummy data')
+                json.set!('projectOutcome4', 'Dummy data')
+                json.set!('projectOutcome5', 'Dummy data')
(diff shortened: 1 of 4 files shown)


app/models/project.rb
```diff
@@ -166,6 +166,67 @@ def to_salesforce_json
...
+                    json.county self.county
+                    json.townCity self.townCity
+                    json.projectPostcode self.postcode
+                end
+                json.set!('yourIdeaProject', self.description)
+                json.set!('projectDifference', self.difference)
+                json.set!('projectCommunity', self.matter)
+                json.set!('projectOrgBestPlace', self.best_placed_description)
+                json.set!('projectAvailable', self.heritage_description)
+                json.set!('projectOutcome1', 'Dummy data')
```
gidsg:
```suggestion
                json.set!('projectOutcome1', self.involvement_description)
```
This is the projectOuctcome1 question

gidsg: Can we update the test to include this field?

gidsg:
Can we update https://github.com/heritagefund/funding-frontend/blob/master/test/models/project_test.rb to check it's pulling through the projectOutcome1 field?

app/models/project.rb
```diff
@@ -166,6 +166,67 @@ def to_salesforce_json
...
+                end
+                json.cashContributions do
+                    json.description 'Dummy data'
+                    json.amount 2000
+                    json.secured 'no'
+                    json.id 'c4237718-ced7-4d03-a95b-1eceaecfdbe0'
+                end
+                json.evidenceOfSupport do
+                    json.description 'Dummy data'
+                end
```
gidsg: I need to update the payload but we decided to remove this block as we will associate the data with the file metadata instead. So this should be removed.

app/models/project.rb
```diff
@@ -166,6 +166,67 @@ def to_salesforce_json
...
+                json.set!('projectOutcome6Checked', true)
+                json.set!('projectOutcome7Checked', true)
+                json.set!('projectOutcome8Checked', true)
+                json.set!('projectOutcome9Checked', true)
+                json.projectCosts do
+                    json.costId 'baa49446-cb70-46c7-ade1-0e17ad450c8a'
+                    json.costType 'new-staff'
+                    json.costDescription 'Dummy data'
+                    json.costAmount '1000'
+                end
```
gidsg: This should be an array of objects rather than an object.

app/models/project.rb
stuartmccoll: Done 👍.

app/models/project.rb
stuartmccoll: Done for all elements that should be arrays containing objects.

PR #102, opened 2020-01-09
Added functionality to project permissions page

stuartmccoll:
This pull request adds functionality to the project permissions page and is based on the latest version of `master`, which includes @PushBackNeil's changes to the conditional radio button functionality.


app/models/project.rb
```diff
@@ -151,6 +188,12 @@ def validate_confirm_declaration?
         validate_confirm_declaration == true
     end
 
+    enum permission_type: {
+        yes: 0,
+        no: 1,
+        not_sure: 2
```
gidsg: I think previously I got a warning when I tried to use `not_sure` as `not_` is reserved prefix, so in that case I changed it to `x_not_sure` https://github.com/heritagefund/funding-frontend/blob/fb0c2a46c9d7df5fccebfc8453a092541d7afe10/app/models/cash_contribution.rb#L7-L8

stuartmccoll: I've resolved this in the same way - all references to `not_sure` are now `x_not_sure`.

PR #104, opened 2020-01-09
Updates to content from Kerry

paulmsmith:


app/views/project/project_matters/show.html.erb
```diff
@@ -13,7 +13,7 @@
 
       <h1 class="govuk-label-wrapper">
         <span class="govuk-caption-xl">About your project</span>
-        <%= f.label :matter, "How does your project matter to you and your community?", class: "govuk-label govuk-label--xl" %>
+        <%= f.label :matter, "Why is your project important to your community?", class: "govuk-label govuk-label--xl" %>
```
stuartmccoll: The page title will also need to change to reflect this - at the moment it's still 'How does your project matter' or 'Errors: How does your project matter?' if there are errors on the page.

paulmsmith: Ah good spot.

paulmsmith: @stuartmccoll Done that now.

PR #110, opened 2020-01-10
Added functionality to project outcomes page

stuartmccoll:
This pull request adds functionality to the project outcomes page.

It'd be nice to split out the conditional text areas into a partial at some point, but we can address that post-MVP.


app/views/project/project_outcomes/show.html.erb
```diff
@@ -0,0 +1,382 @@
...
+
+          </div>
+
+        </div>
+
+        <div class="govuk-checkboxes__item">
+
+          <%= f.check_box :outcome_9,
+                          {
+                              id: "project-conditional-outcome-9-conditional",
```
gidsg: This ID should match the label, currently it appears they do not match i.e. `project-conditional-outcome-9-conditional` v.s. `project_outcome_9` let me know if I've missed something.

app/views/project/project_outcomes/show.html.erb
```diff
@@ -0,0 +1,382 @@
...
+          <%= f.check_box :outcome_2,
+                          {
+                              id: "project-conditional-outcome-2-conditional",
+                              class: "govuk-checkboxes__input",
+                              "data-aria-controls" => "conditional-outcome-2-conditional"
+                          },
+                          "true",
+                          "false"
+          %>
+          <%= f.label :project_outcome_2, "Heritage will be in a better condition", class: "govuk-label govuk-checkboxes__label" %>
```
gidsg: we need the the label ID to match the check box ID for accessibility. So e.g. in this case it should be `project-conditional-outcome-2-conditional`.

PR #112, opened 2020-01-10
make amount field required on client side

gidsg:
fixes exception in case cost amount is empty. (allows for empty amount but we should restrict this by other validations) CC @Ellayanor


app/helpers/project/project_costs_helper.rb
```diff
@@ -1,5 +1,5 @@
 module Project::ProjectCostsHelper
   def calculate_total(project_costs)
-    project_costs.select(:amount).map(&:amount).reduce(:+)
+    project_costs.select(:amount).map(&:amount).compact.reduce(:+)
```
gidsg: this stops it attempting to add nil values.

PR #115, opened 2020-01-10
Added functionality to capital works page

stuartmccoll:
This pull request adds functionality to the project capital works page.

Note that, like the other file upload pages, I haven't tested the non-JS journey. I think we should pick this up separately.


app/controllers/project/project_capital_works_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+  include ProjectContext
+
+  def update
+
+    logger.debug "Updating capital works for project ID: #{@project.id}"
+
+    @project.validate_capital_work = true
+
+    if params[:project].present?
+      @project.validate_capital_work_file = true if params[:project][:capital_work] == "true"
```
gidsg: Quite possibly missing something here, but it looks like we're setting this value to `true` twice? Can one of these be deleted or am I missing something?

app/views/project/project_capital_works/show.html.erb
```diff
@@ -0,0 +1,148 @@
...
+                  <th scope="col" class="govuk-table__header">Condition Survey File</th>
+
+                </tr>
+              </thead>
+              <tbody class="govuk-table__body">
+              <tr>
+                <td>
+                  <%# TODO: Link to download path (currently not working due to PaaS permissions %>
+                  <a href="#">
+                    <%= @project.capital_work_file.blob.filename %>
```
gidsg: This link should work now, as we have resolved the TODO. You can use a link something like: `<%=link_to(@project.capital_work_file.blob.filename, rails_blob_path(project.capital_work_file, disposition: "attachment")) %>`

app/controllers/project/project_capital_works_controller.rb
stuartmccoll: This line is to validate that the capital work file is present, which we only want to do if the `capital_work` value passed in is true, i.e. a user has selected the 'Yes, my project involves capital work' option.

Line 8 is to validate that the value passed in the `capital_work` param is either `true` or `false`, as per the column on the database.

app/views/project/project_capital_works/show.html.erb
stuartmccoll: Done 👍.

app/views/project/project_capital_works/show.html.erb
```diff
@@ -0,0 +1,148 @@
...
+        </div>
+
+        <div class="govuk-radios__item">
+
+          <%= f.radio_button :capital_work,
+                             "true",
+                             class: "govuk-radios__input",
+                             "data-aria-controls" => "conditional-project_capital_work_true"
+          %>
+          <%= f.label :capital_work, "Yes, capital work is part of my project", class: "govuk-label govuk-radios__label" %>
```
gidsg: this label does not match the associated button, I think it should be
```suggestion
          <%= f.label :capital_work_true, "Yes, capital work is part of my project", class: "govuk-label govuk-radios__label" %>
```

app/views/project/project_capital_works/show.html.erb
```diff
@@ -0,0 +1,148 @@
...
+        By capital work we mean repair and conservation works, new building work,
+        digitisation of objects, or the conservation of artefacts and objects.
+      </span>
+
+      <div class="govuk-radios govuk-radios--conditional" data-module="govuk-radios">
+
+        <div class="govuk-radios__item">
+
+          <%= f.radio_button :capital_work, "false", class: "govuk-radios__input" %>
+          <%= f.label :capital_work, "No, capital work is not part of my project", class: "govuk-label govuk-radios__label" %>
```
gidsg: I think label does not match the radio button here.
```suggestion
          <%= f.label :capital_work_false, "No, capital work is not part of my project", class: "govuk-label govuk-radios__label" %>
```

gidsg:
When I try to upload a file on this branch, when it gets to the end of the upload the page hangs. The way we have done this on the other page is to do a normal server side redirect in the controller which rails does after the XHR request but doesn't seem to be happening in this case (at least for me)
Screenshot 2020-01-13 at 09 33 17

gidsg:
> When I try to upload a file on this branch, when it gets to the end of the upload the page hangs. The way we have done this on the other page is to do a normal server side redirect in the controller which rails does after the XHR request but doesn't seem to be happening in this case (at least for me)
> Screenshot 2020-01-13 at 09 33 17

OK I think this is a broder issue I have raised as #124

PR #119, opened 2020-01-10
Add partnership declaration

gidsg:
CC @paulmsmith there is a TODO you could help with here to show hide based on the radio button
@stuartmccoll perhaps you can show me the pattern you use for the server side validation.


app/views/project/project_declaration/show_declaration.html.erb
```diff
@@ -177,7 +177,30 @@
...
+          <%= f.radio_button(:is_partnership, 'false', class: 'govuk-radios__input') %>
+          <%= f.label(:is_partnership_false, 'No', class: 'govuk-label govuk-radios__label') %>
+        </div>
+      </div>
+    </fieldset>
+  </div>
+  <div class="govuk-form-group">
+
+      <%= f.label(:partnership_details, 'Please tell us the organisation(s) you will be delivering your project in partnership with and provide a contact for each partner organisation (name and position).', class: 'govuk-label') %>
+    <%# TODO: Conditionally show this if user selec†s Yes option %>
```
stuartmccoll: Have a look at '[Conditionally revealing content](https://design-system.service.gov.uk/components/radios/)' for radio buttons from the GOV.UK Design System, for the markup to do this. I'd have added it as a code suggestion here, but it's a bunch of extra classes, etc!

app/views/project/project_declaration/show_declaration.html.erb
```diff
@@ -177,7 +177,30 @@
 
   </ul>
 
-  <%# TODO: Include additional partnership fields %>
+  <div class="govuk-form-group">
+    <fieldset class="govuk-fieldset">
+      <span class="govuk-label">
+      Are you applying on behalf of a partnership?
+    </span>
+      <%# TODO: Add validation for this field being set  %>
```
stuartmccoll: In terms of validation, we'd want a `validate_is_partnership?` model method, with a `attr_accessor` for `validate_is_partnership`. Then an extra validation line on the model of `validates :is_partnership, presence: true, if: :validate_is_partnership?`. That should be enough to validate that one of the radio buttons have been selected. Might be worth adding in the extra `validates_inclusion_of` line similar to what I've done with the capital works page to ensure that not only is there a value being passed through in the form parameters, but that it's also either `true` or `false`.

Validation for the conditional field would then be done similar to how we've done it for project permissions. Check if the `is_partnership` parameter is `true`, set `validate_is_partnership_description` to `true` in the controller, and then wire up validations in the model in the same way that we've done for project permissions - checking length if that's something we want to do, or just checking that the field has been passed with some kind of value in it.

gidsg:
@stuartmccoll I haven't managed to get the server-side validation to work, so would it be OK to leave this as a TODO for now and maybe you can walk through it with me when you have some time?

PR #121, opened 2020-01-10
Feature/project dashboard design

paulmsmith:
Adds a little layout and typography niceness to the basic project dashboard.

image


app/views/devise/registrations/edit.html.erb
```diff
@@ -1,6 +1,6 @@
 <% content_for :page_title, "Your account" %>
 
-<p class="govuk-body"><%= link_to "Back", :back, class: "govuk-back-link" %></p>
+<p class="govuk-body"><%= link_to "Back", "javascript:history.back();", class: "govuk-back-link" %></p>
```
paulmsmith: Added this because I noticed the back link wasn't working for some reason. There might be rails way to fix this rather than the javascript history link which works but could also be troublesome.

stuartmccoll: Happy to use this as a starting point, let's see if any issues occur with it.

PR #126, opened 2020-01-13
updates the design for evidence of support page

paulmsmith:
Adding this pull request but need some help @gidsg @stuartmccoll .

I've made the design changes but it looks like the ERB needs updating. Instead of looping out a new table row with each item, it seems there are two separate loops for the descriptions and filenames. Can this be made more like the costs table please?

### DIFF
--- a/app/views/project/project_support_evidence/project_support_evidence.html.erb
+++ b/app/views/project/project_support_evidence/project_support_evidence.html.erb
@@ -26,7 +26,7 @@
               <%= description %>
           </td>
           <td class="govuk-table__cell">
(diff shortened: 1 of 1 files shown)


gidsg:
@paulmsmith because of the validations in place currently and the database structure it is not trivial to loop them together as we cannot guarantee that both the description and file will be present. I suggest we stick with it as-is for user testing and we can raise an issue to re-visit the database structure.

gidsg:
...OK scratch that, I think with the client side validation the approach I have pushed now will work fine. (It may behave strangely for previously incomplete records though).

PR #129, opened 2020-01-14
Handle ajax errors and display client side error messages

PushBackNeil:


app/views/project/project_support_evidence/project_support_evidence.html.erb
```diff
@@ -3,6 +3,20 @@
     <div class="govuk-grid-column-full">
       <span class="govuk-caption-xl">Support for your project</span>
       <h1 class="govuk-heading-xl">Evidence of support</h1>
+      <div id="error-summary-container" class="govuk-visually-hidden">
```
gidsg: It would be better if this wasn't in the DOM unless there is an error as I think it would show for screen reader users. So I think we should use an approach closer to https://github.com/heritagefund/funding-frontend/pull/115/files#diff-885e36ab3c77dac55d0a35e91a1ac8df

app/views/project/project_support_evidence/project_support_evidence.html.erb
```diff
@@ -1,108 +1,109 @@
...
+            <thead class="govuk-table__head">
+            <tr class="govuk-table__row">
+              <th scope="col" class="govuk-table__header">Description</th>
+              <th scope="col" class="govuk-table__header">Files</th>
+            </tr>
+            </thead>
+            <tbody class="govuk-table__body">
+            <tr>
+              <td>
+                <% @project.evidence_description.each do |e| %>
```
gidsg: This erroneously restored the previous way of building the table, which did not have proper table mark-up. I don't think this needs to change as part of this PR so can we revert these changes to the table?

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_01_13_130313) do
+ActiveRecord::Schema.define(version: 2020_01_09_142055) do
```
gidsg: Can we revert these changes? I don't think this PR should alter the schema at all

PushBackNeil: Sorry @gidsg I must stop using git commit -a!

app/views/project/project_support_evidence/project_support_evidence.html.erb
PushBackNeil: The error messages are now injected and removed dynamically similar to what Stuart did.

app/views/project/project_support_evidence/project_support_evidence.html.erb
PushBackNeil: Are there must have been a merge to master whilst I was doing this. I have merged master into this branch and it now has the new styling.

db/schema.rb
stuartmccoll: Reverted.

app/views/project/project_support_evidence/project_support_evidence.html.erb
stuartmccoll: Reverted.

PR #133, opened 2020-01-15
Added migrations and model for non-cash contributions

stuartmccoll:
This pull request adds migrations for the non-cash contributions table and the relevant model, and amends the `Projects` model to include a reference to the `NonCashContributions` model.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_01_10_151233) do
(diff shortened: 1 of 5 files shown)


db/schema.rb
```diff
@@ -108,6 +117,7 @@
     t.integer "permission_type"
     t.text "permission_description"
     t.boolean "capital_work"
+    t.string "capital_work_supporting_document"
```
gidsg: Looks like this is a stray addition to the schema? I suggest doing a `rake db:reset` to remove any things like this.

stuartmccoll: Ah - I'd done a `rails db:reset` prior to this, does that do something different?

PR #136, opened 2020-01-16
Functionality for non-cash contributions

stuartmccoll:
This pull request adds functionality and validation to the non-cash contributions page of the service journey.

I think there has to be a better way of re-rendering the page and retaining the values when validation has failed rather than using `flash`. I'll raise an issue to revisit this post-MVP.

I've also re-used the code from the error summary partial - it'll need updating to be able to handle nested attributes too. Rather than get bogged down with that here, I'll raise an issue to revisit.

Note that the design isn't in place for the table - one for @paulmsmith to look at separately.


app/controllers/project/project_non_cash_contributions_controller.rb
```diff
@@ -0,0 +1,42 @@
+class Project::ProjectNonCashContributionsController < ApplicationController
+  include ProjectContext
+
+  def update
+
+    # Empty flash values to ensure that we don't redisplay them unnecessarily
+    flash[:description_value] = ""
```
gidsg: are we sure this is being used, it doesn't seem to be referenced anywhere else?

app/views/project/project_non_cash_contributions/show.html.erb
```diff
@@ -0,0 +1,194 @@
...
+  <div class="nlhf-summary__body">
+
+    <% if @project.non_cash_contributions.first.try(:id).nil? %>
+
+      <h3 class="govuk-heading-m govuk-!-margin-bottom-0">You have not added any non-cash contributions</h3>
+
+    <% else %>
+
+      <% @project.non_cash_contributions.filter { |ncc| ncc.id.present? }.each do |ncc| %>
+
```
gidsg: Should we add a TODO here to do a proper design for this section?

app/controllers/project/project_non_cash_contributions_controller.rb
stuartmccoll: Good spot! It was supposed to be `flash[:description]` - fixed in the latest commit.

app/views/project/project_non_cash_contributions/show.html.erb
stuartmccoll: Added 👍.

app/controllers/project/project_non_cash_contributions_controller.rb
```diff
@@ -4,7 +4,7 @@ class Project::ProjectNonCashContributionsController < ApplicationController
   def update
 
     # Empty flash values to ensure that we don't redisplay them unnecessarily
-    flash[:description_value] = ""
+    flash[:description] = ""
```
gidsg: I wonder if using [flash.now](https://guides.rubyonrails.org/action_controller_overview.html#flash-now) would negate the need to instantiate these as empty? But not critical for this PR.

stuartmccoll: I tried a couple of different ways on one of the previous controllers and nothing seemed to work consistently except instantiating them as empty. Happy to revisit at some point, ideally I'd like to get this working without the use of `flash` at all. It feels like it should be possible 🙂.

stuartmccoll: Raised #137 to revisit the use of `flash`.

PR #138, opened 2020-01-17
Wired up cash contributions question page

stuartmccoll:
This pull request contains functionality to wire up the 'Are you getting cash contributions?' page, a conditional page which will either redirect a user to the cash contributions page or the non-cash contributions page, depending on their answer.


app/views/project/project_costs/show.html.erb
```diff
@@ -235,7 +235,11 @@
...
-  <%# TODO: Button should redirect to next route %>
-  <button class="govuk-button" role="button" aria-label="save and continue button">Save and continue</button>
+<hr class="govuk-section-break--l govuk-section-break--visible">
+
+<p class="govuk-body">If you have finished adding all your costs</p>
+
+<a href="<%= "#{three_to_ten_k_project_cash_contributions_question_get_url}" %>" role="button" draggable="false" class="govuk-button"
+   data-module="govuk-button" aria-label="Continue button">
+  Save and continue
+</a>
```
gidsg: I wonder if from an an accessibility point of view whether it's better to mark this up as an actual form with a `<button>` with a form get method input?

stuartmccoll: I'll raise an issue to investigate, I think we're using this in quite a few places at the minute.

stuartmccoll: Raised #140 to investigate this.

PR #139, opened 2020-01-17
Wired grant request page into service flow

stuartmccoll:
This pull request wires the 'Your grant request' page into the service journey.


app/controllers/project/project_grant_request_controller.rb
```diff
@@ -1,12 +1,63 @@
...
+      logger.debug "Retrieved project cost ID #{project_cost.id}, amount #{project_cost.amount} " +
+                       "for project ID: #{@project.id}"
+
+      @total_project_cost += project_cost.amount
+
+      logger.debug "Running total project cost for project ID: #{@project.id} is: #{@total_project_cost}"
+
+    end
+
+    logger.debug "Calculated total project cost of #{@total_project_cost} for project ID: #{@project.id}"
```
gidsg: I think we could just use this existing helper

app/controllers/project/project_grant_request_controller.rb
```diff
@@ -1,12 +1,63 @@
...
+                       "for project ID: #{@project.id}"
+
+      @total_cash_contributions += cash_contribution.amount
+
+      logger.debug "Running total cash contributions for project ID: #{@project.id} is: #{@total_cash_contributions}"
+
+    end
+
+    logger.debug "Calculated total cash contributions of #{@total_cash_contributions} for project ID: #{@project.id}"
+
```
gidsg: I think we could just use or adapt this existing helper: https://github.com/heritagefund/funding-frontend/blob/650c5ab44f9a6ed12c3940a91671c2cc5df87092/app/helpers/project/project_costs_helper.rb#L2-L4

app/views/project/project_grant_request/show.html.erb
```diff
@@ -0,0 +1,78 @@
+<% content_for :page_title, "Your grant request" %>
+
+<h1 class="govuk-heading-xl">
+  <span class="govuk-caption-xl">About your project</span>
+  Your grant request
+</h1>
+
+<div class="govuk-panel govuk-panel--confirmation govuk-!-margin-bottom-7">
+  <h1 class="govuk-panel__title">
+    £<%= '%.2f' % @final_grant_amount %>
```
gidsg: We can use a helper which includes the pound symbol and formatting the number e.g.
```suggestion
    <%= number_to_currency(@final_grant_amount, strip_insignificant_zeros: true) %>
```

app/views/project/project_grant_request/show.html.erb
```diff
@@ -0,0 +1,78 @@
...
+
+      <h2 class="govuk-heading-s govuk-!-margin-bottom-3">Your cash contributions</h2>
+
+      <p class="govuk-body">This is the amount of money you told us you will receive for your project that does not include your grant request</p>
+
+    </div>
+
+    <div class="nlhf-grant-summary__amount">
+
+      <p>£<%= @total_cash_contributions %></p>
```
gidsg:
```suggestion
      <p><%= number_to_currency(@total_cash_contributions, strip_insignificant_zeros: true) %></p>
```

app/controllers/project/project_grant_request_controller.rb
```diff
@@ -1,12 +1,63 @@
...
-    @total_project_cost = 100
-    @cash_contributions = 50
-    @final_grant_amount = @total_project_cost - @cash_contributions
+  include ProjectContext
+
+  def show
+
+    calculate_total_project_cost
+
+    calculate_total_cash_contributions
```
gidsg: Can we remove these, as they are not assigned I don't _think_ they are doing anything?

stuartmccoll: They're calling the two `private` methods, but I'll look at adapting the existing helper.

app/controllers/project/project_grant_request_controller.rb
```diff
@@ -3,61 +3,11 @@ class Project::ProjectGrantRequestController < ApplicationController
 
   def show
 
-    calculate_total_project_cost
-
-    calculate_total_cash_contributions
+    @total_project_cost = helpers.calculate_total(@project.project_costs)
+    @total_cash_contributions = helpers.calculate_total(@project.cash_contributions)
```
gidsg: I don't _think_ these need to be instance variables as we don't appear to be using them in the view? Only to calculate the value below?

stuartmccoll: They get used in the 'Here's how we calculated this' section.

app/views/project/project_grant_request/show.html.erb
```diff
@@ -0,0 +1,79 @@
...
+
+      <h2 class="govuk-heading-s govuk-!-margin-bottom-3">Your total project costs</h2>
+
+      <p class="govuk-body">This is what you told us the total amount of money your whole project will cost</p>
+
+    </div>
+
+    <div class="nlhf-grant-summary__amount">
+
+      <p>£<%= @total_project_cost %></p>
```
gidsg: we can use the currency helper here.
```suggestion
      <p><%= number_to_currency(@total_project_cost, strip_insignificant_zeros: true) %></p>
```

app/views/project/project_grant_request/show.html.erb
```diff
@@ -0,0 +1,79 @@
...
+
+<h1 class="govuk-heading-xl">
+  <span class="govuk-caption-xl">About your project</span>
+  Your grant request
+</h1>
+
+<div class="govuk-panel govuk-panel--confirmation govuk-!-margin-bottom-7">
+  <h1 class="govuk-panel__title">
+    <%= number_to_currency(@final_grant_amount, strip_insignificant_zeros: true) %>
+  </h1>
```
gidsg: we should not have multiple H1s on the page. Maybe make these both H2s?

PR #141, opened 2020-01-17
Added validation to project volunteers page

stuartmccoll:
This pull request adds validation to the project volunteers page. Note that this is done in the same way that we have done non-cash contributions, using `flash` to populate form values when we re-render upon validation failure, which we've said that we'll revisit when we have time.

I've also left in the `TODO` around splitting the summary of errors out into a partial, or refactoring the existing partial to be able to accommodate nested forms.

### DIFF
--- a/app/views/project/project_volunteers/show.html.erb
+++ b/app/views/project/project_volunteers/show.html.erb
@@ -107,6 +107,7 @@
 
               <%= v.label :description, class: 'govuk-label' %>
               <%= v.text_area :description,
+                              required: true,
                               class: "govuk-textarea #{'govuk-textarea--error' if @project.errors['volunteers.description'].any?} govuk-js-character-count",
                               rows: 5,
                               value: "#{flash[:description] if flash[:description].present?}"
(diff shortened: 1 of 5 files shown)


app/views/project/project_volunteers/show.html.erb
```diff
@@ -1,85 +1,135 @@
...
-          <div class="govuk-form-group">
+          <div class="govuk-form-group <%= "govuk-form-group--error" if @project.errors['volunteers.hours'].any? %>">
+
+            <%= render partial: "partials/form_input_errors",
+                       locals: {form_object: @project,
+                                input_field_id: :'volunteers.hours'} if @project.errors['volunteers.hours'].any? %>
+
             <%= v.label :hours, class: 'govuk-label' %>
-            <%= v.number_field :hours, class: 'govuk-input govuk-input govuk-input--width-10' %>
+            <%= v.number_field :hours,
```
gidsg: Can we add `required: true` to these fields to do client-side validation?

stuartmccoll: Done :+1:.

PR #143, opened 2020-01-17
Added additional content to account registration page

PushBackNeil:


app/controllers/application_controller.rb
```diff
@@ -1,2 +1,9 @@
 class ApplicationController < ActionController::Base
+  before_action :configure_permitted_parameters, if: :devise_controller?
+
+  protected
+
+  def configure_permitted_parameters
+    devise_parameter_sanitizer.permit(:sign_up, keys: [:date_of_birth, :name, :line1, :line2, :line3, :townCity, :county, :postcode, :phone_number ])
+  end
```
gidsg: I _think_ we can revert these changes as they should only be required if you are updating these fields when creating the account.

app/controllers/dashboard_controller.rb
```diff
@@ -3,7 +3,31 @@ class DashboardController < ApplicationController
...
+    user_details_fields_presence.push(
+        (
+        current_user.line1.present? ||
+            current_user.townCity.present? ||
+            current_user.county.present? ||
+            current_user.postcode.present?
+        )
+    )
+
+    return user_details_fields_presence.include?(false) ? false : true
```
gidsg:
```suggestion
    return user_details_fields_presence.all?
```
I think this is a more concise way to test for truthyness.

app/controllers/dashboard_controller.rb
```diff
@@ -3,7 +3,31 @@ class DashboardController < ApplicationController
...
+  private
+  def user_details_complete
+
+    user_details_fields_presence = []
+
+    user_details_fields_presence.push(current_user.name.present?)
+    user_details_fields_presence.push(current_user.date_of_birth.present?)
+    user_details_fields_presence.push(
+        (
+        current_user.line1.present? ||
```
gidsg: This will return true if _any_ of these is present. Was that the intent?

stuartmccoll: It was, but the more I think about it, I prefer the way I was originally doing it, which was to check that they were _all_ present, as we're mandating that each of these has to be entered. What do you think?

app/controllers/dashboard_controller.rb
```diff
@@ -3,7 +3,31 @@ class DashboardController < ApplicationController
 
   def show
 
-    @projects = Project.all.where(user_id: current_user.id)
+    if user_details_complete == false
```
gidsg:
```suggestion
    unless user_details_complete
```

app/controllers/dashboard_controller.rb
```diff
@@ -3,7 +3,31 @@ class DashboardController < ApplicationController
...
+      redirect_to user_details_path
+    else
+      @projects = Project.all.where(user_id: current_user.id)
```
gidsg:
```suggestion
      @projects = current_user.projects
```

app/controllers/dashboard_controller.rb
```diff
@@ -3,7 +3,31 @@ class DashboardController < ApplicationController
...
+  end
+
+  private
+  def user_details_complete
+
+    user_details_fields_presence = []
+
+    user_details_fields_presence.push(current_user.name.present?)
+    user_details_fields_presence.push(current_user.date_of_birth.present?)
+    user_details_fields_presence.push(
```
gidsg: Can we do a single select query to get all these fields and then check? Else I think it will do a select per request which is less performant.

app/views/devise/registrations/new.html.erb
```diff
@@ -29,7 +29,7 @@
 <h3 class="govuk-heading-m">Problems creating an account</h3>
 
 <%- if devise_mapping.confirmable? && controller_name != 'confirmations' %>
-  <%= link_to "Didn't receive confirmation instructions?", new_confirmation_path(resource_name) %><br />
+  <%= link_to "Didn't receive confirmation instructions?", new_confirmation_path(resource_name) %><br/>
 <% end %>
 
 <%- if controller_name != 'sessions' %>
```
gidsg: This looks like just whitespace changes. Can we revert?

app/views/user/details/show.html.erb
```diff
@@ -0,0 +1,146 @@
...
+
+  </fieldset>
+
+  <p class="govuk-body-s">
+    The personal information we collect from you will be shared with fraud prevention agencies who will
+    use it to prevent fraud and money-laundering and to verify your identity. If fraud is detected, you
+    could be refused a grant or loan. Further details of how your information will be used by us and these
+    fraud prevention agencies, and your data protection rights, can be found by contacting
+    <%=mail_to 'fair.processing@heritagefund.org.uk' %>.
+  </p>
```
gidsg: Seems like we are displaying this message on 2 different pages, is that intentional?

stuartmccoll: It is - I'm expecting Paul to come in after me and clean up the actual page content.

db/schema.rb
```diff
@@ -175,6 +175,15 @@
...
     t.string "unconfirmed_email"
+    t.date "date_of_birth"
+    t.string "name"
+    t.string "line1"
+    t.string "line2"
+    t.string "line3"
+    t.string "townCity"
+    t.string "county"
+    t.string "postcode"
+    t.string "phone_number"
```
gidsg: I think the timestamp on the schema needs to change else the app won't know to run the migration. You could do this manually by updating the timestamp to be after the last one run, or automatically by resetting your DB then running the migration.

app/controllers/dashboard_controller.rb
gidsg: Yeah, I don't mind, depends on the intent. If we want the simplest proxy, maybe we just check for `line1` and trust the other validation works. Else use `&&`.

app/controllers/dashboard_controller.rb
stuartmccoll: I think that `current_user` is set when the page loads and contains all attributes, so I don't think it's actually running further database queries. The only thing that gets logged out when hitting the dashboard controller is the following line, which is logged on all routes requiring user authentication:

```
User Load (0.4ms)  SELECT "users".* FROM "users" WHERE "users"."id" = $1 ORDER BY "users"."id" ASC LIMIT $2  [["id", 3], ["LIMIT", 1]]
```

db/schema.rb
stuartmccoll: Resetting the database and running the migration wasn't doing anything, so I've manually cranked the migration timestamp and timestamp in the schema. Should be okay now.

app/controllers/dashboard_controller.rb
stuartmccoll: I've changed it back to `&&`, which I had in there originally.

PR #144, opened 2020-01-17
Added validation to project costs page

stuartmccoll:
This pull request adds validation to the project costs page in the same way that we have done for the non-cash contributions page.

Note that this pull request also moves the `details` component to the bottom of the page, and removes the alphabetised ordering from the `select` component, as discussed as an output of the most recent round of user research.


app/views/project/project_costs/show.html.erb
```diff
@@ -1,240 +1,378 @@
...
+
+    <% end %>
+
+  </div>
+
+</section>
+
+<h1 class="govuk-heading-l govuk-!-margin-bottom-6">Add a cost</h1>
+
+<h1 class="govuk-heading-s">Add a new cost for each project cost.</h1>
```
gidsg: we should only have one H1 on the page. Can we make these `<h2>`?

stuartmccoll: Done 👍.

PR #147, opened 2020-01-17
Add Salesforce active job

gidsg:
This adds an [active job ](https://guides.rubyonrails.org/active_job_basics.html) to submit a project to Salesforce. It is modelled like this so we can optionally run it asynchronously, on a separate worker with retries etc.
Note this is currently not wired-up to any controller but you can trigger it from the command line to try it out
e.g. 
```
rails runner "ApplicationToSalesforceJob.perform_now(Project.find('e4da66c3-5cdf-4ab5-b507-d287e0d19557'))"
```
In the process of doing this I had to make a few app level changes which are documented in the commit messages.


app/models/project.rb
```diff
@@ -245,11 +246,11 @@ def validate_file_attached(field, error_msg)
...
 
-        word_count = self.public_send(field).split(' ').count
+        word_count = self.public_send(field)&.split(' ')&.count
 
         logger.debug "#{field} word count is #{word_count}"
 
-        if word_count > max_length
+        if word_count && word_count > max_length
             self.errors.add(field, error_msg)
         end
```
gidsg: These changes allow this validator to still work in the case of a `nil` value.

app/mailers/notify_mailer.rb
```diff
@@ -54,4 +54,14 @@ def unlock_instructions(record, token, opts = {})
     )
   end
 
+  #  @param [Project] project
```
stuartmccoll: I'm assuming this line isn't doing anything and can be removed?

gidsg: This annotation helps IDEs know what the argument type is to provide Intellisense.

stuartmccoll: 👍

PR #152, opened 2020-01-21
Added an extra guard to the set_project method

### LINKED ISSUE #151: Only retrieve projects if they are linked to the current user
At the moment the `set_project` method within the `ProjectContext` class will retrieve a `project` object based on the `:project_id` parameter passed in a URL, and then set this as the `@project` instance variable.

We should only retrieve a `project` object if it belongs to the `current_user`, otherwise we should return a user to the projects dashboard.

stuartmccoll:
When setting the `@projec`t instance variable, we now check to ensure that the project object being retrieved is linked to the id of the current user. If it is, we proceed to set the `@project` instance variable as normal, otherwise we redirect the user to the projects dashboard.

Closes #151


stuartmccoll:
> This is the sort of thing we should have a test for! But maybe a TODO would be fine for now.
> 
> Do we need to do a similar check around organisation routes?

Yeah, I figured we'd pick it up when we come to our 'test everything' sprint. I'll add a `TODO` as well.

I'll raise an issue for organisations, because it should be there too.

PR #153, opened 2020-01-21
Replace setup-ruby with use-ruby-action

stuartmccoll:
The [setup-ruby](https://github.com/actions/setup-ruby/) GitHub Action only includes one minor release of Ruby per version. This has recently been updated to bump `2.6.x` to `2.6.5`, which means that `2.6.4` (our Ruby version) is no longer available. Rather than reporting this error, the Action seems to default to then installing `2.7.0`, which causes the `Gemfile` installation to fail.

The [use-ruby-action](https://github.com/eregon/use-ruby-action) provides the ability to specify any version of Ruby, which it will then download. We should use this in the short term until [setup-ruby](https://github.com/actions/setup-ruby/) has been updated to accommodate other Ruby versions.


stuartmccoll:
Closing, as we've discussed bumping the Ruby version to match that used in `setup-ruby` instead.

PR #154, opened 2020-01-21
Updated Ruby version

stuartmccoll:


.github/workflows/continuous-integration-workflow.yml
```diff
@@ -23,9 +23,9 @@ jobs:
     - uses: actions/setup-node@v1
       with:
         node-version: '12.x'
-    - uses: actions/setup-ruby@v1.0.0
+    - uses: actions/setup-ruby@v1.1.1
```
gidsg: Does this work If we change it to just `@v1`? I'm not sure how the versioning works on Github actions?

stuartmccoll: Let's see! We'll squash commits after, as there's loads of rubbish commits in here.

stuartmccoll: It does :+1:.

PR #158, opened 2020-01-21
Added validation to cash contributions page

stuartmccoll:
This pull request adds validation to the Cash contributions page.

With the JavaScript validation, there's still a fair amount of refactoring that can be done - at some stage I'd like to move it all into a single place, and have it abstracted enough that each view can use it independently. I'd like to get this in for each page as necessary first, and then pick up refactoring as a single task later on.


app/models/cash_contribution.rb
```diff
@@ -2,6 +2,44 @@ class CashContribution < ApplicationRecord
...
+
+    word_count = self.public_send(field)&.split(' ')&.count
+
+    logger.debug "#{field} word count is #{word_count}"
+
+    if word_count && word_count > max_length
+      self.errors.add(field, error_msg)
+    end
+
+  end
```
gidsg: I think these are copy pasted from other controllers? Can we add a TODO to move to shared validators (e.g. using a concern)?

stuartmccoll: I've gone one better and moved them to a shared `concern` now 🙂.

PR #159, opened 2020-01-22
Replaced set_organisation helper use with a controller context

### LINKED ISSUE #155: Only retrieve organisations if they are linked to the current user
Similar to #151 for projects, we should only retrieve an organisation object if it belongs to the current_user, otherwise we should return a user to the projects dashboard.

stuartmccoll:
Similar to how we replaced the `set_project` usage for project controllers, this pull request does the same for organisation controllers.

As well as that, it adds extra checks to ensure that the organisation which is being accessed is related to the current user.

Closes #155

### DIFF
--- a/app/controllers/concerns/organisation_context.rb
+++ b/app/controllers/concerns/organisation_context.rb
@@ -1,0 +1,29 @@
+module OrganisationContext
+  extend ActiveSupport::Concern
+  included do
+    before_action :authenticate_user!, :set_organisation
+  end
+
+  # This method retrieves an organisation object based on the id
+  # found in the URL parameters and the current authenticated
+  # user's id, setting it as an instance variable for use in
+  # organisation-related controllers.
+  #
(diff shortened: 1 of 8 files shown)


app/controllers/concerns/organisation_context.rb
```diff
@@ -0,0 +1,29 @@
...
+  # user's id, setting it as an instance variable for use in
+  # organisation-related controllers.
+  #
+  # If no organisation object matching the parameters is found,
+  # then the user is redirected to the projects dashboard.
+  #
+  # TODO: Write tests for this.
+  def set_organisation
+
+    authorised = false
```
gidsg: Can we do this without mutating the `authorised` value? As it makes it a bit harder to reason about what's going on. Perhaps in a separate action e.g. `check_authorisation`?

e.g. 
`redirect_to :authenticated_root unless params[:organisation_id] == current_user.organisation&.id`

PR #160, opened 2020-01-22
Split evidence of support into own model and added validation

stuartmccoll:
This pull request splits supporting evidence into it's own model and adds validation to the relevant page.

Note that it's the same validation used on similar pages, I'll revisit DRY-ing this and making it reusable and slightly more robust once the entire journey is wired up and in place.

### DIFF
--- a/app/models/evidence_of_support.rb
+++ b/app/models/evidence_of_support.rb
@@ -1,0 +1,30 @@
+class EvidenceOfSupport < ApplicationRecord
+  include GenericValidator
+
+  # This overrides Rails attempting to pluralise the model name
+  self.table_name = "evidence_of_support"
+
+  self.implicit_order_column = "created_at"
+
+  belongs_to :project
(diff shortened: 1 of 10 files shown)


app/models/evidence_of_support.rb
```diff
@@ -0,0 +1,30 @@
+class EvidenceOfSupport < ApplicationRecord
+  include GenericValidator
+
+  # This overrides Rails attempting to pluralise the model name
+  self.table_name = "evidence_of_support"
+
+  self.implicit_order_column = "created_at"
+
+  belongs_to :project
+  has_many_attached :evidence_of_support_files
```
gidsg: I think as it is designed currently, each evidence of support has one file attached? 
If so, we can change this and not use the `first` later on to find the file.

stuartmccoll: You're right, this was a hangover from the previous way of modelling. I'll change this now.

PR #169, opened 2020-01-23
Updated initial organisation validation

### LINKED ISSUE #167: user can skip to project creation without completing all organisation fields
# Steps to reproduce
* Register a new user
* Click through to start now
* Log out
* Log back in

# Actual Results
User is redirected to set project details

# Expected Results
User cannot proceed to project creation until organisation fields are populated

stuartmccoll:
This pull request extends the existing validation that takes place to determine whether or not a user should be directed to the organisation or project sections of the service. 

Previously, we were only checking to see whether a user had a related organisation, directing them to the project section if so or to the organisation section if not.

Now, we check to see whether the user has a related organisation and whether or not this organisation has the mandatory organisation attributes populated. If not, we direct the user to the organisation section of the service. If the user does not have an organisation, we still direct them into the organisation section of the service. If the user has an organisation and this organisation has the necessary mandatory attributes populated then we direct them to the project section of the service.

As well as this, this pull request fixes a bug where Turbolinks was causing a project to be created twice from the 'Start a project' page, and fixes a bug which was causing a 500 error due to a mis-named route.

Closes #167

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -10,21 +10,22 @@ class HomeController < ApplicationController
 
     if current_user.organisation
 
-      logger.info "Existing organisation #{current_user.organisation.id} found for #{current_user.id.to_s}"
+      logger.info "Existing organisation #{current_user.organisation.id} found for #{current_user.id}"
 
-      @start_now_link_value = three_to_ten_k_project_create_path
+      @start_now_link_value = complete_organisation_details? ? three_to_ten_k_project_create_path :
+                                  organisation_type_path(current_user.organisation.id)
 
     else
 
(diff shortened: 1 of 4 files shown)


app/controllers/home_controller.rb
```diff
@@ -36,4 +37,22 @@ def show
 
   end
 
+  private
+  # Checks for the presence of mandatory organisation parameters,
+  # returning false if any are not present and true if all are
+  # present
+  def complete_organisation_details?
+
+    return [
```
gidsg: Very minor but I don't think we need `return` here. Ruby will by default return the last statement.
```suggestion
    [
```

PR #177, opened 2020-01-24
Added validation to confirm presence of project costs

stuartmccoll:
This pull request adds a validation check to confirm the presence of associated project costs, as per [this](https://trello.com/c/kIXpjTKo) Trello card.

### DIFF
--- a/app/controllers/project/project_costs_controller.rb
+++ b/app/controllers/project/project_costs_controller.rb
@@ -2,6 +2,28 @@ class Project::ProjectCostsController < ApplicationController
   include ProjectContext
   before_action :remove_flash_values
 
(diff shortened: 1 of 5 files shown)


stuartmccoll:
> I am slightly confused where in here we are actually checking for the existence of a project cost? But I trust it's in there somewhere!

`validates :project_costs, presence: true, if: :validate_has_associated_project_costs?`

A few different StackOverflow threads seemed to think that this was the best way of doing it rather than adding a custom validator to check that the count was over 0, etc. Do you want to test it your end to confirm it works?

gidsg:
@stuartmccoll that's fine, I just was not previously aware of this particular rails magic 🎩

PR #178, opened 2020-01-24
Added running of Salesforce job to declaration controller

stuartmccoll:
This pull request adds functionality to run the Salesforce application submission job if a user has confirmed that they have read the declaration.

### DIFF
--- a/app/controllers/project/project_declaration_controller.rb
+++ b/app/controllers/project/project_declaration_controller.rb
@@ -13,6 +13,8 @@ class Project::ProjectDeclarationController < ApplicationController
 
       logger.debug "Finished updating declaration confirmation for project ID: #{@project.id}"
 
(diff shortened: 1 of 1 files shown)


app/controllers/project/project_declaration_controller.rb
```diff
@@ -13,6 +13,8 @@ def update_confirm_declaration
 
       logger.debug "Finished updating declaration confirmation for project ID: #{@project.id}"
 
+      ApplicationToSalesforceJob.perform_later(@project)
```
gidsg: Can we log an info line before this e.g. _triggering salesforce job for Application ID_ in case we need to debug issues it is usually better to have more logging.

stuartmccoll: :+1:

PR #179, opened 2020-01-24
Made cash contributions secured value display dynamically

stuartmccoll:
Previously we weren't displaying the actual `:secured` value for cash contributions in the table on this page. This pull request addresses this - removing the `x_` from the value for 'Not sure', humanising each value and then replacing `Yes ` with `Yes, ` for readability.

As well, this fixes a bug where cash contribution calculations included 'hidden' fields which weren't being displayed in the table because they did not have descriptions.

The `link_to` line was causing 500 errors, which has been fixed also.

### DIFF
--- a/app/views/project/project_cash_contribution/show.html.erb
+++ b/app/views/project/project_cash_contribution/show.html.erb
@@ -24,7 +24,7 @@
         </tr>
         </thead>
         <tbody class="govuk-table__body">
-        <% @project.cash_contributions.filter { |cc| cc.id.present? }.each_with_index do |cc, index| %>
+        <% @project.cash_contributions.filter { |cc| cc.id.present? }.each do |cc| %>
           <tr class="govuk-table__row">
             <td class="govuk-table__cell nlhf-table__cell govuk-!-width-one-half nlhf-table__cell--text">
               <%= cc.description %>
--- a/app/views/project/project_cash_contribution/show.html.erb
+++ b/app/views/project/project_cash_contribution/show.html.erb
@@ -24,17 +24,16 @@
         </tr>
         </thead>
         <tbody class="govuk-table__body">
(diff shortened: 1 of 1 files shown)


app/views/project/project_cash_contribution/show.html.erb
```diff
@@ -24,17 +24,16 @@
...
-        <% @project.cash_contributions.filter { |cc| cc.description.present? }.each_with_index do |cc, index| %>
+        <% @project.cash_contributions.filter { |cc| cc.id.present? }.each_with_index do |cc, index| %>
           <tr class="govuk-table__row">
             <td class="govuk-table__cell nlhf-table__cell govuk-!-width-one-half nlhf-table__cell--text">
               <%= cc.description %>
             </td>
             <td class="govuk-table__cell">
-              <%# TODO: Add human friendly value of the choosen radio box "supports" %>
-              Value here
+              <%= cc.secured&.gsub('x_', '').humanize.gsub('Yes ', 'Yes, ') %>
```
gidsg: `.gsub('Yes ', 'Yes, ')`
This seems to be doing a replacement to the same value unless I am missing something?

app/views/project/project_cash_contribution/show.html.erb
```diff
@@ -24,17 +24,16 @@
         </tr>
         </thead>
         <tbody class="govuk-table__body">
-        <% @project.cash_contributions.filter { |cc| cc.description.present? }.each_with_index do |cc, index| %>
+        <% @project.cash_contributions.filter { |cc| cc.id.present? }.each_with_index do |cc, index| %>
```
gidsg:
```suggestion
        <% @project.cash_contributions.filter { |cc| cc.id.present? }.each do |cc| %>
```
I don't _think_ we are using the index here so I think it can be safely removed, unless I'm missing something.

PR #180, opened 2020-01-24
Added contacts details to Salesforce payload

stuartmccoll:
This pull request adds functionality to retrieve the correct data for the `mainContact` fields of the Salesforce payload, and also contains updates to the relevant test and fixtures.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -312,7 +312,7 @@ class Project < ApplicationRecord
                     json.endDate self.end_date
                 end
                 json.set!('mainContactName', self.user.name)
-                json.set!('mainContactDateOfBirth', self.user.date_of_birth)
+                json.set!('mainContactDateOfBirth', self.user.date_of_birth.to_s)
                 json.set!('mainContactEmail', self.user.email)
                 json.set!('mainContactPhone', self.user.phone_number)
                 json.mainContactAddress do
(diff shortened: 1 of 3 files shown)


app/models/project.rb
```diff
@@ -311,15 +311,17 @@ def to_salesforce_json
                     json.startDate self.start_date
                     json.endDate self.end_date
                 end
-                #TODO: Replace this dummy data with real data
-                json.set!('mainContactName', "Jane Doe")
-                json.set!('mainContactDateOfBirth', "1975-10-12")
-                json.set!('mainContactEmail', "test@example.net")
+                json.set!('mainContactName', self.user.name)
+                json.set!('mainContactDateOfBirth', self.user.date_of_birth)
```
gidsg: Does `date_of_birth` stringify in the required format by default?

stuartmccoll: Good spot. Fixed.

app/models/project.rb
```diff
@@ -312,7 +312,7 @@ def to_salesforce_json
                     json.endDate self.end_date
                 end
                 json.set!('mainContactName', self.user.name)
-                json.set!('mainContactDateOfBirth', self.user.date_of_birth)
+                json.set!('mainContactDateOfBirth', self.user.date_of_birth.to_s)
```
gidsg: sorry, this seems to be the default behaviour so I don't think it is required.

PR #181, opened 2020-01-24
adds in logic to view to validate grant request

paulmsmith:
This is probably really noddy but does what we intend for the grant request page.

- Value greater than funding programme maximum will not let users continue and prompts to view costs and contributions
- Will not let users continue if their grant request is invalid due to cash contributions being higher than project costs.

@stuartmccoll or @gidsg - This will likely need refactoring but hopefully you can see how I've structured the content, mark-up depending on those scenarios.

### DIFF
--- a/app/controllers/project/project_grant_request_controller.rb
+++ b/app/controllers/project/project_grant_request_controller.rb
@@ -5,8 +5,11 @@ class Project::ProjectGrantRequestController < ApplicationController
 
     @total_project_cost = helpers.calculate_total(@project.project_costs).to_i
     @total_cash_contributions = helpers.calculate_total(@project.cash_contributions).to_i
-
     @final_grant_amount = @total_project_cost - @total_cash_contributions
+    @maximum_request_value = 10000.to_i
+    @costs_greater_than_contributions = @final_grant_amount.positive? ? true : false
+    @grant_request_less_than_funding_band = @final_grant_amount <= @maximum_request_value ? true : false
+    @grant_request_is_valid = @costs_greater_than_contributions && @grant_request_less_than_funding_band ? true : false
 
   end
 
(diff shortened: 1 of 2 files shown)


app/controllers/project/project_grant_request_controller.rb
```diff
@@ -5,8 +5,11 @@ def show
 
     @total_project_cost = helpers.calculate_total(@project.project_costs).to_i
     @total_cash_contributions = helpers.calculate_total(@project.cash_contributions).to_i
-
     @final_grant_amount = @total_project_cost - @total_cash_contributions
+    @maximum_request_value = 10000.to_i
+    @costs_greater_than_contributions = @final_grant_amount.positive? ? true : false
```
gidsg:
```suggestion
    @costs_greater_than_contributions = @final_grant_amount.positive? 
```
I think we can remove this as it already returns a boolean

app/controllers/project/project_grant_request_controller.rb
```diff
@@ -5,8 +5,11 @@ def show
...
+    @grant_request_less_than_funding_band = @final_grant_amount <= @maximum_request_value ? true : false
```
gidsg:
```suggestion
    @grant_request_less_than_funding_band = @final_grant_amount <= @maximum_request_value 
```
ibid

app/controllers/project/project_grant_request_controller.rb
```diff
@@ -5,8 +5,11 @@ def show
...
+    @grant_request_is_valid = @costs_greater_than_contributions && @grant_request_less_than_funding_band ? true : false
```
gidsg:
```suggestion
    @grant_request_is_valid = @costs_greater_than_contributions && @grant_request_less_than_funding_band
```

PR #182, opened 2020-01-24
Implemented deletion of project costs

stuartmccoll:
This pull request adds the ability to delete a project cost from the 'Project costs' page.

As well, this pull request contains the validation for the project cost description, which was previously found to be missing.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -124,6 +124,9 @@ Rails.application.routes.draw do
 
       get ':project_id/costs', to: 'project_costs#show', as: :project_costs
       put ':project_id/costs', to: 'project_costs#update'
+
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -124,6 +124,9 @@
 
       get ':project_id/costs', to: 'project_costs#show', as: :project_costs
       put ':project_id/costs', to: 'project_costs#update'
+
+      get ':project_id/costs/:project_cost_id', to: 'project_costs#delete', as: :cost_delete
```
gidsg: Can we do this using `delete` HTTP action and do a form delete on the page? This would make it more RESTy.

stuartmccoll: 👍

PR #184, opened 2020-01-24
adds tracking snippet to layout

paulmsmith:
image

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -31,18 +31,6 @@
     </main>
   </div>
   <%= javascript_pack_tag 'application', 'data-turbolinks-suppress-warning':'true', 'data-turbolinks-eval': "false", 'data-turbolinks-track': 'reload' %>
-  <% if Rails.env.production? %>
-    <%# Hotjar tracking tag  %>
-    <script>
-        (function(h,o,t,j,a,r){
-            h.hj=h.hj||function(){(h.hj.q=h.hj.q||[]).push(arguments)};
-            h._hjSettings={hjid:<%= Rails.configuration.x.hotjar.id %>,hjsv:<%= Rails.configuration.x.hotjar.snippet_version %>};
-            a=o.getElementsByTagName('head')[0];
-            r=o.createElement('script');r.async=1;
-            r.src=t+h._hjSettings.hjid+j+h._hjSettings.hjsv;
-            a.appendChild(r);
-        })(window,document,'https://static.hotjar.com/c/hotjar-','.js?sv=');
-    </script>
-  <% end %>
+  <%= render "partials/tracking_tags" if Rails.env.production? %>
 </body>
 </html>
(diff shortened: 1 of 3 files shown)


app/views/layouts/application.html.erb
```diff
@@ -31,5 +31,18 @@
     </main>
   </div>
   <%= javascript_pack_tag 'application', 'data-turbolinks-suppress-warning':'true', 'data-turbolinks-eval': "false", 'data-turbolinks-track': 'reload' %>
+  <% if Rails.env.production? %>
+    <%# Hotjar tracking tag  %>
+    <script>
```
gidsg: Can we put `defer` attribute on this so it is non blocking?

app/views/layouts/application.html.erb
```diff
@@ -31,5 +31,18 @@
...
+        (function(h,o,t,j,a,r){
+            h.hj=h.hj||function(){(h.hj.q=h.hj.q||[]).push(arguments)};
+            h._hjSettings={hjid:<%= Rails.configuration.x.hotjar.id %>,hjsv:<%= Rails.configuration.x.hotjar.snippet_version %>};
+            a=o.getElementsByTagName('head')[0];
+            r=o.createElement('script');r.async=1;
+            r.src=t+h._hjSettings.hjid+j+h._hjSettings.hjsv;
+            a.appendChild(r);
+        })(window,document,'https://static.hotjar.com/c/hotjar-','.js?sv=');
+    </script>
+  <% end %>
```
gidsg: Can we move this into a partial and then conditionally include it?

config/environments/production.rb
```diff
@@ -125,5 +125,7 @@
   config.x.salesforce.client_secret = ENV.fetch("SALESFORCE_CLIENT_SECRET")
   # TODO: update this with production host once credentials available
   config.x.salesforce.host = "test.salesforce.com"
+  config.x.hotjar.id = ENV.fetch("HOTJAR_ID")
+  config.x.hotjar.snippet_version = ENV.fetch("HOTJAR_SNIPPET_VERSION")
```
gidsg: I would just hardcode this rather than including it in environment as it's not a secret.

stuartmccoll: I was going with it here for ease of changing it - seemed like it'd be quicker for us to make an environment variable change if necessary than pushing up a code change.

gidsg: Not sure that's true as we need to do a deploy anyway to change an env var without downtime, so my preference if it's not a secret (which this isn't as it's exposed on the client-side), would be to hard-code it in config.

stuartmccoll: Ah! I was thinking about this yesterday and it seems like it's something we're not likely to need to change anyway - once we've got a Hotter key, I doubt we're likely to invalidate it and need another one. I'll strip this out and hardcode it in a partial.

app/views/layouts/application.html.erb
paulmsmith: We can do. It needs to be on every page so I just added it straight to layout. :)

config/environments/production.rb
gidsg: @stuartmccoll doesn't like this change was pushed, can you push it up then I'll approve the PR?

stuartmccoll: I haven't done it yet - it's on my list next 👍.

paulmsmith: I've created the partial now @stuartmccoll if you pull down the latest from this branch

PR #194, opened 2020-01-27
Organisation numbers tidying

stuartmccoll:
This pull request contains general tidying and slight refactoring of the controller and view for the organisation numbers page of the service.


app/views/organisation/numbers/show.html.erb
```diff
@@ -1,27 +1,45 @@
...
+%>
 
-<%= render partial: "partials/summary_errors", locals: {
-        form_object: @organisation,
-        first_form_element: :organisation_company_number
-} if @organisation.errors.any? %>
+<h1 class="govuk-heading-xl" aria-label="Heading">About your organisation</h1>
 
-<%= form_with model: @organisation, url: organisation_numbers_path, method: :put do |f|  %>
+<%= form_with model: @organisation, url: :organisation_numbers, method: :put, local: true do |f|  %>
```
gidsg: What is the thinking behind removing `_path` here?

app/views/organisation/numbers/show.html.erb
```diff
@@ -31,16 +49,32 @@
...
+    <%=
+      render partial: "partials/form_input_errors",
+             locals: {
+                 form_object: @organisation,
+                 input_field_id: :charity_number
+             } if @organisation.errors[:charity_number].any?
+    %>
+
+    <%=
+      f.number_field :charity_number,
```
gidsg: Should this be a text field instead now?

app/views/organisation/numbers/show.html.erb
stuartmccoll: The `:` means that it isn't needed - `:organisation_number` will resolve to the same thing as `organisation_numbers_path`.

app/views/organisation/numbers/show.html.erb
stuartmccoll: Yep, it is in `master` - this branch needs updating.

stuartmccoll: Merged `master` into this branch, so this has been resolved.

PR #195, opened 2020-01-27
General tidying of organisation mission section

stuartmccoll:
This pull request contains some tidying and slight refactoring of the controller and view for the 'Organisation mission' section of the service.

### DIFF
--- a/app/views/organisation/mission/show.html.erb
+++ b/app/views/organisation/mission/show.html.erb
@@ -1,14 +1,27 @@
-<h1 class="govuk-heading-l" aria-label="Heading">About your organisation</h1>
+<%=
+  render partial: "partials/page_title",
+         locals: {
+             model_object: @organisation,
+             page_title: "Tell us about the mission, or objectives, of your organisation"
+         }
+%>
 
-<%= render partial: "partials/summary_errors", locals: {
-    form_object: @organisation,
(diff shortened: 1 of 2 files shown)


app/views/organisation/mission/show.html.erb
```diff
@@ -19,39 +32,103 @@
...
-          <%= f.check_box :mission, {multiple: true, id: :organisation_young_people_led, class: "govuk-checkboxes__input"}, "young_people_led", nil  %>
+          <%=
+            f.check_box :mission,
+                        {
+                            multiple: true,
+                            id: :organisation_young_people_led,
+                            class: "govuk-checkboxes__input"
+                        },
+                        "young_people_led",
+                        nil
```
gidsg: do we need this `nil` argument? Does it not default to `nil` without this?

stuartmccoll: Without the `nil` value here, submitting the form results in the following being submitted, as an example when only 'Black or minority ethnic led' has been checked:

```
 "organisation"=>{"mission"=>["0", "black_or_minority_ethnic_led", "0", "0", "0", "0"]}
```

The `nil` prevents it from sending in the spurious `"0"` values when a checkbox hasn't been checked.

PR #210, opened 2020-01-29
Add support pages to service

stuartmccoll:
This pull request implements support pages in the same style as GOV.UK Notify, as suggested for our initial implementation.

To be iterated upon later by @paulmsmith.


app/controllers/support_controller.rb
```diff
@@ -0,0 +1,164 @@
...
+
+    unless params[:support_question_or_feedback_message].present?
+      flash[:errors][:support_question_or_feedback_message] = "Enter a message"
+    else
+      flash[:support_question_or_feedback_message] = params[:support_question_or_feedback_message]
+    end
+
+    flash[:support_question_or_feedback_name] = params[:support_question_or_feedback_name] if
+        params[:support_question_or_feedback_name].present?
+    flash[:support_question_or_feedback_email] = params[:support_question_or_feedback_email] if
```
gidsg: Can we look into using a validator rather than manually manipulating the flash message? OK for this to be a todo for now.

stuartmccoll: Raised #212 to pick this up at a later date.

PR #218, opened 2020-01-29
Feature/support page

paulmsmith:
Adds content to the support page and updates site-wide feedback link

### DIFF
--- a/app/views/support/show.html.erb
+++ b/app/views/support/show.html.erb
@@ -10,7 +10,7 @@
          } if flash[:errors]
 %>
 
-<%= form_tag :support, local: true do |f| %>
+<%= form_tag :support, class: "govuk-!-margin-bottom-3", local: true do |f| %>
 
   <div class="govuk-form-group <%= "govuk-form-group--error" if flash[:errors] %>">
 
@@ -19,7 +19,7 @@
       <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
 
         <h1 class="govuk-fieldset__heading">
-          Suppport
+          Support
         </h1>
 
       </legend>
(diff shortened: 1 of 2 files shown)


app/views/support/show.html.erb
```diff
@@ -56,6 +56,30 @@
...
-<% end %>
\ No newline at end of file
+<% end %>
+
+    <hr class="govuk-section-break govuk-!-margin-bottom-7 govuk-section-break--visible">
+    
+    <h2 class="govuk-heading-m">Help and guidance</h2>
+
+    <ul class="govuk-list govuk-list--bullet govuk-!-margin-bottom-6">
+      <li><a href="https://www.heritagefund.org.uk/funding/check-what-we-fund" class="govuk-link">Check what we fund</a></li>
```
stuartmccoll: I don't think this needs the `class` attribute, as the other links in this list seem to pick up the same styling by default.

app/views/support/show.html.erb
```diff
@@ -56,6 +56,30 @@
...
+    <h3 class="govuk-heading-s">Phone</h3>
+    <p class="govuk-body">020 7591 6044</p>
+
+    <h3 class="govuk-heading-s">Text phone</h3>
+    <p class="govuk-body">020 7591 6255</p>
+
+    <p class="govuk-body">Our office hours are 9:00am to 5:00pm, Monday to Friday.</p>
+    
+    <h3 class="govuk-heading-s">By email</h3>
+    <p class="govuk-body"><a href="mailto:enquire@heritagefund.org.uk">enquire@heritagefund.org.uk</a></p>
```
gidsg:
```suggestion
    <p class="govuk-body"><%=mail_to 'enquire@heritagefund.org.uk' %></p>
```
we can use a helper here.

PR #223, opened 2020-01-30
Added flag to enable non-JS journey on Capital Work page

stuartmccoll:
With JavaScript disabled, submitting the form on the 'Will capital work be part of your project?' page would throw an `ActionController::InvalidAuthenticityToken` exception. By setting a `no_js` variable inside a `noscript` element, we can then use this flag to determine whether or not to set the `remote` value of the form to `true` or `false`.

I'm not sure whether this is best practice - there might be a built-in Rails way of determining this, but I couldn't find anything when researching.

### DIFF
--- a/app/views/project/project_capital_works/show.html.erb
+++ b/app/views/project/project_capital_works/show.html.erb
@@ -1,13 +1,15 @@
 <% content_for :page_title, @project.errors.any? ? "Error: Will capital work be part of your project?" : "Will capital work be part of your project?" %>
 
+<noscript><% no_js = true %></noscript>
+
 <div id="summary-errors"></div>
 
 <%= render partial: "partials/summary_errors", locals: {
     form_object: @project,
-    first_form_element: :project_capital_work_no
+    first_form_element: :project_capital_work_false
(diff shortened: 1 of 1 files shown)


app/views/project/project_capital_works/show.html.erb
```diff
@@ -1,13 +1,15 @@
 <% content_for :page_title, @project.errors.any? ? "Error: Will capital work be part of your project?" : "Will capital work be part of your project?" %>
 
+<noscript><% no_js = true %></noscript>
```
gidsg: Can we use the presence of `js-enabled` on the body to enable this instead? i.e. we already append this class in:
https://github.com/heritagefund/funding-frontend/blob/9cb2f9dfdc851351a7d673fb00e4b9897f376180/app/views/layouts/application.html.erb#L14 ?

paulmsmith: 👍

stuartmccoll: Created #224 to revisit this when we have time.

PR #230, opened 2020-01-30
Replaces HTML5 email field type with text type for the input (across the app)

paulmsmith:
@stuartmccoll or @gidsg I can't see a problem with this but there is nothing in the rails/ruby code that is somehow expecting that field type is there? (I doubt it but thought I'd ask)


stuartmccoll:
My only concern with this is that removing this validation means that we aren't validating the data being entered into some of the email fields in the service. There might be existing server-side validation built into the relevant Devise controllers, but we definitely aren't doing it on the legal signatories or support pages, for example. Are we happy to go with this initially and then backfill the validation later on?

paulmsmith:
> My only concern with this is that removing this validation means that we aren't validating the data being entered into some of the email fields in the service. There might be existing server-side validation built into the relevant Devise controllers, but we definitely aren't doing it on the legal signatories or support pages, for example. Are we happy to go with this initially and then backfill the validation later on?

I personally would be happy to back fill as we can always phone applicants and have other safe guards in place. "Email didn't arrive", etc.

PR #231, opened 2020-01-30
add sentry client-side code

gidsg:


app/javascript/sentry/index.js
```diff
@@ -0,0 +1,5 @@
+import * as Sentry from '@sentry/browser';
+
+Sentry.init({
+    dsn: 'https://d1df242da6ee423b8f7d588f2730a633@sentry.io/2095964'
```
gidsg: we hardcode this because it is built at build time and available on the client-side.

PR #250, opened 2020-01-31
Converted organisation company number from integer to string

stuartmccoll:
This pull request converts the organisation company number attribute from an integer to a string.


gidsg:
We should remove the redundant string conversions here: https://github.com/heritagefund/funding-frontend/blob/176a7f679b6e355dd7770cc4a9f15f9f0fd0e2be/app/models/project.rb#L398-L399

PR #253, opened 2020-01-31
Added status code to SalesforceApexError raise

stuartmccoll:


app/jobs/application_to_salesforce_job.rb
```diff
@@ -42,7 +42,7 @@ def perform(project)
         ApplicationAttachmentsToSalesforceJob.perform_later(salesforce_case_id, cc, :cash_contribution_evidence_files, cc.description)
       end
     else
-      raise SalesforceApexError.new("Failure response from Salesforce when POSTing project ID: #{project.id}")
+      raise SalesforceApexError.new("Failure response from Salesforce when POSTing project ID: #{project.id}, status code: #{@response.status}")
```
gidsg:
```suggestion
      raise SalesforceApexError.new("Failure response from Salesforce when POSTing project ID: #{project.id}, status code: #{@response&.status}")
```
safe navigation in case nil we don't want to throw.

PR #254, opened 2020-02-03
Added email validation to legal signatory model and controller

stuartmccoll:
This looks like the simplest way to add this in for now - although it will allow email addresses through in the format of `name@mailserver`, for example.

Note that it is no longer necessary to validate `presence: true` for this field, as it already triggers as part of the email validation. Leaving it in results in two error messages being displayed when no email address is added.


gidsg:
Can we do this in a shared validator and apply to all email fields? e.g. similar to that shown in: https://guides.rubyonrails.org/active_record_validations.html#custom-validators

PR #259, opened 2020-02-04
Disables 'turbolinks' on all pages

paulmsmith:
See here in documentation: https://github.com/turbolinks/turbolinks#disabling-turbolinks-on-specific-links


gidsg:
@paulmsmith can you double check functionality still works with this change e.g. add costs, file upload. Then I am happy to approve and we can raise a follow up PR to remove other associated code.

stuartmccoll:
I've merged `master` back into this branch and run through a couple of pages within the service to ensure that functionality isn't broken - supporting evidence, project costs, cash contributions. I think it's still worth doing a double check and running this on somebody else's machine.

PR #261, opened 2020-02-04
Initial iteration of JavaScript refactoring

stuartmccoll:
This pull request encompasses the first piece of work to refactor our JavaScript error-handling into a single place. In this initial iteration, I have moved the creation/amendment of the 'summary of errors' page element (see [here](https://design-system.service.gov.uk/components/error-summary/) in the GOV.UK Design System) into a single place.


app/javascript/packs/application.js
```diff
@@ -7,9 +7,10 @@ require("turbolinks").start();
 require("@rails/activestorage").start();
 require("channels");
 require("govuk");
-require("../sentry/index")
+require("../sentry/index");
 require("core-js/stable");
 require("regenerator-runtime/runtime");
+require("common");
```
gidsg: can we give this a more descriptive name e.g. `error-handler`?

app/javascript/common/index.js
```diff
@@ -0,0 +1,155 @@
...
+  Function used to create the skeleton code for the GOV.UK-styled
+  summary errors element
+ */
+window.createSummaryErrorsSkeleton = function() {
+
+    var summaryErrorsElement = document.getElementById("summary-errors");
+
+    addSummaryErrorsClassAndAttributes(summaryErrorsElement);
+
+    if (summaryErrorsElement.innerHTML.includes("h2") === false) {
```
gidsg: I am a bit confused by the negative checks we are doing here? Can we make them positive checks instead? Could we use a switch statement instead?

app/javascript/common/index.js
```diff
@@ -0,0 +1,155 @@
+/*
+  Function used to create the skeleton code for the GOV.UK-styled
+  summary errors element
+ */
+window.createSummaryErrorsSkeleton = function() {
```
gidsg: rather than attaching directly to `window` is it an improvement if we call this function directly from the initialiser? Something a bit like the pattern here? https://github.com/heritagefund/funding-frontend/blob/71d2669b4c5caa9bf368f5666024a82f62bf32e3/app/javascript/packs/application.js#L7

stuartmccoll: Definitely! I don't think I saw this in a single piece of Rails documentation - lots of 'workarounds', but no 'you just need to add `.start()`'!

gidsg: I am not sure how magical this is, i.e. it looks like they call `window` within `.start()` but I guess the entry point is clearer.

app/javascript/common/index.js
stuartmccoll: We're checking to ensure that the summary errors component does not already contain our error message (because we don't want to add it more than once), and that the error attribute isn't the same as the name of the nested model (because Rails adds a, in our case spurious, extra key to the errors hash of `nested model name` with a message of `is invalid`.

I can update this code with a comment. Not sure how well it lends itself to flipping the logic - thoughts?

gidsg: I think adding a comment would be fine.

app/javascript/error-handling/index.js
```diff
@@ -0,0 +1,155 @@
...
+  Function used to create the skeleton code for the GOV.UK-styled
+  summary errors element
+ */
+createSummaryErrorsSkeleton = function() {
+
+    var summaryErrorsElement = document.getElementById("summary-errors");
+
+    addSummaryErrorsClassAndAttributes(summaryErrorsElement);
+
+    if (summaryErrorsElement.innerHTML.includes("h2") === false) {
```
gidsg:
```suggestion
    if (!summaryErrorsElement.innerHTML.includes("h2")) {
```
I _think_ this is will be functionally equivalent and a bit more concise?

PR #263, opened 2020-02-05
Removal of individual data-turbolinks attributes

stuartmccoll:
As we're now setting this site-wide, there is no need to include individual `data-turbolinks` attributes across the application. This pull request removes each of these individual uses.

### DIFF
--- a/app/views/project/project_capital_works/show.html.erb
+++ b/app/views/project/project_capital_works/show.html.erb
@@ -9,7 +9,7 @@
     first_form_element: :project_capital_work_false
 } if @project.errors.any? %>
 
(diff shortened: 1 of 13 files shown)


app/views/project/project_capital_works/show.html.erb
```diff
@@ -9,7 +9,7 @@
     first_form_element: :project_capital_work_false
 } if @project.errors.any? %>
 
-<%= form_for @project, url: three_to_ten_k_project_capital_works_put_path, method: :put, remote: no_js ? false : true do |f| %>
+<%= form_with model: @project, url: three_to_ten_k_project_capital_works_put_path, method: :put, remote: no_js ? false : true do |f| %>
```
stuartmccoll: Note here, that because we were previously using `form_for` the form was never sending remotely, so it was not triggering the AJAX handling.

PR #267, opened 2020-02-06
Updates for clearer pilot information

paulmsmith:
New banner component
Updated content on sign-in page

### DIFF
--- a/app/views/devise/sessions/new.html.erb
+++ b/app/views/devise/sessions/new.html.erb
@@ -2,11 +2,21 @@
 
 <%= render "devise/shared/error_messages", resource: resource %>
 
-<h2 class="govuk-heading-l">Sign in</h2>
 
-<p class="govuk-body govuk-!-margin-bottom-6">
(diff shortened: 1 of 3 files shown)


paulmsmith:
image

app/views/devise/sessions/new.html.erb
```diff
@@ -2,11 +2,21 @@
 
 <%= render "devise/shared/error_messages", resource: resource %>
 
-<h2 class="govuk-heading-l">Sign in</h2>
 
-<p class="govuk-body govuk-!-margin-bottom-6">
-  This is a new application form. If you haven't logged in here before, you might not have an account with us. You will need to <%= link_to "create a new account", new_registration_path(resource_name) %> to apply.
-</p>
+<div class="nlhf-banner">
```
gidsg: we could move this banner block to a partial?

paulmsmith: Eventually yeh. I agree. I was being quick today.

PR #270, opened 2020-02-07
Fixes for org_type and application submitted bugs

stuartmccoll:
This pull request contains a fix for the missing organisation type value in Salesforce.

The values we store for an organisation's `org_type` attribute do not match those in the Salesforce schema for `organisationType`. I have added an additional method to the `project` model that will convert the values to a matching schema value. I have tested this in Salesforce for each value and confirmed that it works as expected.

Whilst testing, I discovered that if the `submitted_on` value for a project is filled before the `redirect_to` the application submitted route happens, then the user will be redirected straight to the project dashboards page, as the check in `set_project` triggers. This pull request also contains a fix for this - only considering `submitted_on` within `set_project` if the `request.path` does not include `/application-submitted`.


app/models/project.rb
```diff
@@ -424,4 +425,28 @@ def to_salesforce_json
             end
         end
     end
+
+    def get_organisation_type_for_salesforce_json
+
+        case self.organisation.org_type
```
gidsg: we should be able directly assign the result of the case statement e.g. 
```
formatted_org_type_value = case self.organisation.org_type
             when 'registered_company', 'community_interest_company'
                  'registered-company-or-community-interest-company'
```
and so on.

app/models/project.rb
```diff
@@ -424,4 +425,28 @@ def to_salesforce_json
             end
         end
     end
+
+    def get_organisation_type_for_salesforce_json
```
gidsg: can we make this a private method?

gidsg:
Can we update the test, to use one of these custom mapped values?

PR #277, opened 2020-02-10
Add feature tests for About You and Postcode Lookup

gidsg:
Adds feature tests for some initial steps in the journey. We should keep these under review as they are slower to run and more tied to the UI but may give us more confidence in the end-to-end service.


spec/factories/users.rb
```diff
@@ -6,7 +6,7 @@
                 strategy: :create
 
     f.confirmed_at { "2020-02-05 12:20:14.391143" }
-    f.email { "test@test.com" }
+    sequence(:email){|n| "user#{n}#{Random.new.random_number}@example.com" }
```
stuartmccoll: Do we gain anything by leaving the `n` in email address? I'm guessing it's in case `Random.new.random_number` manages to generate the same random number when creating two different users?

spec/features/address_lookup_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    click_button 'Find address'
+    expect(page).to have_current_path('/user/address/results')
+    select('4 Barons Court Road', from: 'address')
+    click_button 'Select'
+    expect(page).to have_current_path('/user/address/show')
+    expect(page).to have_field('Address line 1', with: '4 Barons Court Road')
+    expect(page).to have_field('Town', with: 'LONDON')
+    expect(page).to have_field('County', with: 'London')
+    expect(page).to have_field('Postcode', with: 'W14 9DT')
+  end
```
stuartmccoll: Is it worth continuing to click the 'Save and continue' button and then checking the `user` attributes to ensure that they've been set?

spec/factories/users.rb
gidsg: The random number was a nasty workaround I added. In theory sequence should be enough but I think I was having issues because we are not cleaning the DB between runs. So I raised an issue for that separately. https://github.com/heritagefund/funding-frontend/issues/278

PR #279, opened 2020-02-11
Updates to worse contrast ratio but is technically the "brand colour"

paulmsmith:
**Current colour and contrast ratio**
image

**"Brand" colour and contrast ratio**
image


paulmsmith:
Thanks @gidsg merging. Brand colour is 'AAA' at least.

PR #284, opened 2020-02-12
Refactoring of legal signatories

stuartmccoll:
This pull request contains refactoring of the organisation legal signatories model, view, and controller. We have refactored the view to use the same nested attributes pattern used elsewhere throughout the service.

Note that a new private method has been included in the controller which will remove a second legal signatory if it has been added with empty attributes. We are only required to validate the second legal signatory if it has been passed through the form with no attributes populated. This is possible, and has been enabled using a lambda in the legal_signatory model. However, when ignoring validation, Rails will attempt to store the record - it's possible to reject a nested model on the parent model, but it isn't possible to specify the parameters for this rejection, i.e. only reject the second nested model passed through. Because of this, we're always choosing not to validate the second nested model if it hasn't been populated, but Rails will then save an almost-empty record to the database (auto-generated attributes are still populated, such as created_at, etc).

The remove_empty_signatory method is therefore necessary to clean up after ourselves when we have chosen not to validate the second empty legal signatory.

We have discussed making neither legal signatory mandatory at this point, depending on how the mandatory version tests. If that is the case, we can refactor this controller and related models again to reject all empty objects, something which isn't possible at the moment, as discussed above.

### DIFF
--- a/app/controllers/organisation/signatories_controller.rb
+++ b/app/controllers/organisation/signatories_controller.rb
@@ -1,90 +1,73 @@
 class Organisation::SignatoriesController < ApplicationController
-  include OrganisationContext
-  before_action :set_legal_signatories
+  include OrganisationContext, ObjectErrorsLogger
+
+  def show
+
+    @organisation.legal_signatories.build unless
+        @organisation.legal_signatories.first.present?
+    @organisation.legal_signatories.build unless
+        @organisation.legal_signatories.second.present?
+
+  end
 
   def update
 
-    logger.debug "Updating legal signatories for organisation ID: #{@organisation.id}"
+    logger.info "Updating legal signatories for organisation ID: " \
+                 "#{@organisation.id}"
 
-    set_signatory_validation_paramaters
+    @organisation.validate_legal_signatories = true
 
-    signatory_validation_statuses = []
+    @organisation.update(organisation_params)
 
-    @legal_signatories.each do |signatory|
+    if @organisation.valid?
 
-      logger.debug "Assigning attributes for legal signatory ID: #{signatory.id}"
+      # See method documentation for description of why this is necessary
+      remove_empty_signatory
 
-      signatory.assign_attributes({
-                           name: params[:legal_signatories][signatory.id][:name],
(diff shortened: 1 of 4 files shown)


app/controllers/organisation/signatories_controller.rb
```diff
@@ -1,90 +1,73 @@
...
 
-      signatory_validation_statuses << signatory.valid?
+    @organisation.update(organisation_params)
 
-    end
+    if @organisation.valid?
 
-    if signatory_validation_statuses.include?(false)
+      # See method documentation for description of why this is necessary
+      remove_empty_signatory
```
gidsg: Could we do this instead in a callback to prevent it writing to the database in the first place i.e. using one of the hooks in https://api.rubyonrails.org/classes/ActiveRecord/Callbacks.html or https://api.rubyonrails.org/classes/AbstractController/Callbacks/ClassMethods.html ?

app/views/organisation/signatories/show.html.erb
```diff
@@ -12,25 +19,32 @@
...
             </li>
-          <% end %>
+        <% end %>
+
+        <% @organisation.legal_signatories.second.errors.each do |attr, msg| %>
+          <li>
+            <a href='#organisation_legal_signatories_attributes_1_<%= attr %>'>
+              <%= msg %>
+            </a>
+          </li>
```
gidsg: could we use `each_with_index` and do this in one block? (Seems like the index number is the only difference?)

app/views/organisation/signatories/show.html.erb
```diff
@@ -39,50 +53,104 @@
...
 
-  <%= submit_tag "Save and continue", {class:'govuk-button', role: 'button', 'aria-label' => 'save and continue button'} %>
+  <%=
+    f.submit "Save and continue",
+             {
+                 class:'govuk-button',
+                 role: 'button',
+                 'aria-label' => 'save and continue button'
+             }
+  %>
```
gidsg: could use `f.save_and_continue` here if `master` is merged in.

PR #288, opened 2020-02-13
add nlhf text field example

gidsg:
Started this yesterday, opening as draft PR so it doesn't get lost, in case it is useful for the component work.
@stuartmccoll there should be all the context here to move the form error messages partial into this component but I didn't get as far as doing it.


stuartmccoll:
I'm going to close this off, can discuss when you're back from leave, but we've written an architectural decision record on this [here](https://github.com/heritagefund/decision-records/blob/master/architecture/001-use-of-actionview-components.md).

PR #290, opened 2020-02-13
Static content pages

paulmsmith:
As part of building out components I needed a place to put examples. There is also a need for us to create some pages of content such as a service specific accessibility statement.

This work is to enable the creation of 'static' pages. Once merged the app will send every url that matches /pages/ to a pages controller. Meaning any of the following would now work:

/pages/about
/pages/home
/pages/features

So long as a file is placed in the `app/views/pages` directory. For example:

A request to `/pages/about` would serve up the contents of `app/views/pages/about.html.erb`

@stuartmccoll I've pieced this together following rails docs. I'm not sure if this needs tests, etc?


paulmsmith:
OK. Having discussed this with @stuartmccoll we've decided to park this for now because it is less flexible should we want to change the URL of some of these content pages.

PR #295, opened 2020-02-19
Added an ActionView Component for Save and continue button

stuartmccoll:


paulmsmith:
The work I've done is based upon the GOVUK [Nunjucks component](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/button/template.njk) with some of the logic placed in the component's 'ActionView::Component' ruby file (in this case: `button_component.rb`)

paulmsmith:
I have not written tests because I'm not sure how, we might need to pair on that @stuartmccoll ? or you could write me an example to follow?

PR #298, opened 2020-02-25
Refactoring of AJAX error handling

stuartmccoll:
This pull request contains refactoring of the AJAX error-handling code being used on pages with file uploads - capital works, evidence of support and cash contributions. Some of the initial work towards this had already been done on a separate feature branch and merged into `master`.

I've tested this in both the non-JavaScript and JavaScript enabled journeys, and both continue to work as they did previously.

I have not pulled Neil's error handling partial into this code - I think that this should be done separately.


stuartmccoll:
> Have you tried your browser stack account yet @stuartmccoll might be worth a look if you've not already?

Good shout. I've only used the most vanilla of vanilla JS, but I'll run it across BrowserStack anyway 👍.

PR #301, opened 2020-02-26
adds custom tracking ability to app

paulmsmith:
Adds the ability to send Hotjar a custom tracking string.


app/views/layouts/application.html.erb
```diff
@@ -56,6 +56,13 @@
   </div>
 </footer>
   <%= javascript_pack_tag 'application', 'data-turbolinks-suppress-warning':'true', 'data-turbolinks-eval': "false", 'data-turbolinks-track': 'reload' %>
-  <%= render "partials/tracking_tags" if Rails.env.production? %>
+  
+  <%# if this is production render the tracking partial %>
+  <%= render partial: "partials/tracking_tags", locals: {
+    # if the layout has been passed content for a custom tracking override pass that into the
+    # partial otherwise pass false
+    tracking_url_path: ((content_for? :tracking_override) ? yield(:tracking_override) : false)
```
stuartmccoll: We can do this without assigning the value to `false`, which should still have the same functionality when it comes to the partial itself:

```suggestion
    tracking_url_path: (yield(:tracking_override) if content_for? :tracking_override)
```

PR #310, opened 2020-02-28
Added integration test for project involvement controller

stuartmccoll:
This pull request adds integration tests for the `project/involvement_controller`.


paulmsmith:
It ran the tests OK but wondered if this message warranted your attention?

stuartmccoll:
@paulmsmith I've got a reminder to look at the cause of that message at some point, but it's not something worth worrying about at the moment - we're very unlikely to be updating the Rails version in the near future 🙂.

PR #328, opened 2020-03-09
Added integration tests for project volunteers controller

stuartmccoll:
This pull request adds integration tests for the `project/volunteers_controller`.

This commit also contains some code cleanup. Adding spacing in the view to increase readability, changing some of the log levels, etc.


app/controllers/project/volunteers_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+  end
+
+  # This method deletes a project volunteer, redirecting back to
+  # :three_to_ten_k_project_volunteers once completed
+  def delete
+
+    logger.info "User has selected to delete volunteer ID: " \
+                "#{params[:volunteer_id]} from project ID: #{@project.id}"
+
+    volunteer = Volunteer.find_by_id(params[:volunteer_id])
```
gidsg: Can we use `find_by` or `find_by!` here depending on the intended failure behaviour? I don't _think_ `find_by_id` exists in current Rails?

stuartmccoll: Much better, that's the functionality I was looking for! We initially had `find(params[:volunteer_id])`, which would throw an exception if it couldn't find it. I didn't realise that `find_by_id` was deprecated, by plain `find_by(id: params[:volunteer_id]` does what we need it to do.

app/controllers/project/volunteers_controller.rb
```diff
@@ -47,7 +47,7 @@ def delete
     logger.info "User has selected to delete volunteer ID: " \
                 "#{params[:volunteer_id]} from project ID: #{@project.id}"
 
-    volunteer = Volunteer.find_by_id(params[:volunteer_id])
+    volunteer = Volunteer.find_by(id: params[:volunteer_id])
```
gidsg: we could use `find` here which uses ID.
```suggestion
    volunteer = Volunteer.find(params[:volunteer_id])
```

app/controllers/project/volunteers_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+    logger.info "User has selected to delete volunteer ID: " \
+                "#{params[:volunteer_id]} from project ID: #{@project.id}"
+
+    volunteer = Volunteer.find_by(id: params[:volunteer_id])
+
+    if volunteer.present?
+
+      logger.info "Deleting volunteer ID: #{volunteer.id}"
+
+      volunteer.destroy if volunteer.project_id == @project.id
```
gidsg: I think we could simplify this by doing the lookup by association rather than an additional if statement e.g. something like
`@project.volunteers.find(params[:volunteer_id]).destroy` 
and remove some of the other checks?

PR #330, opened 2020-03-10
Added integration tests for project non-cash contributions controller

stuartmccoll:
This pull request adds integration tests for the `projects/non_cash_contributions_controller`.

This commit also contains some code cleanup. Adding spacing in the relevant views to increase readability, changing some of the log levels, etc.


app/controllers/project/non_cash_contributions_controller.rb
```diff
@@ -0,0 +1,131 @@
...
+  end
+
+  # This method adds a non-cash contribution to a project, redirecting back to
+  # :three_to_ten_k_project_non_cash_contributions_get if successful and
+  # re-rendering :show method if unsuccessful
+  def update
+
+    # Empty flash values to ensure that we don't redisplay them unnecessarily
+    flash[:description] = ""
+    flash[:amount] = ""
```
gidsg: Have we tried using [`flash.now`](https://guides.rubyonrails.org/action_controller_overview.html#flash-now) when setting errors  rather than setting the empty value to clear the flash?
_this is speculative but worth a try_

stuartmccoll: I _think_ I did, but it's been a while since I've touched the `flash` stuff. I'd like to revisit the use of `flash` as a whole for storing form input values on controllers like this - I thought we had a GitHub issue raised for it already, but we didn't, so I've just created #332 .

PR #338, opened 2020-03-10
upgrade to ruby 2.7

gidsg:
This upgrades to Ruby 2.7 which includes some nice new features. https://www.ruby-lang.org/en/news/2019/12/25/ruby-2-7-0-released/
In order to make this work I had to update ActionView which works using the latest version.

This is an option for us now as the latest Cloudfoundry buildpack supports this version.

In my testing the app itself works fine although Rails and other libraries are showing many deprecation warnings relating to future changes.
We could choose to silence these to avoid gunking up logs i.e. https://dev.to/andrewmcodes/hiding-ruby-2-7-deprecation-warnings-in-rails-6-2mil
since they are not related to our code.

Locally this worked for me once I had updated all the things e.g. _rbenv_, IntelliJ.


stuartmccoll:
👍

Looks good to me - I've pulled down and run a quick regression locally, and confirmed that the tests are running as expected.

PR #341, opened 2020-03-11
add cfignore to remove cached dependencies

gidsg:
speculative fix to stop cloudfoundry using dependencies cached from CI.


gidsg:
going to merge this to get build pipeline to run, will revert if it doesn't work.

PR #354, opened 2020-03-18
add Location controller tests

gidsg:
https://trello.com/c/MJskmrjr/1230-write-integration-tests-for-the-project-location-controller
Note this doesn't add tests for the address lookup functionality as it's very similar to that done for organisation but let me know if you feel strongly this needs to be added.


spec/controllers/project/project_location_controller_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+  end
+
+
+
+  describe "PUT #update" do
+
+    it "should re-render the page with error if no location option selected" do
+      put :update,
+        params: {project_id: project.id, project: {}}
```
stuartmccoll: In our other controller tests, we also test that the same re-rendering and error messages are present if no `project` param is passed in the `params` at all.

spec/controllers/project/project_location_controller_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      ).to eq(I18n.t("activerecord.errors.models.project.attributes.same_location.blank"))
+    end
+
+    it "updates project address to match organisation address if same location selected" do
+      subject.current_user.update(organisation: organisation)
+      put :update,
+          params: {project_id: project.id, project: {same_location: 'yes'}}
+      expect(response).to have_http_status(:redirect)
+      expect(response).to redirect_to(:three_to_ten_k_project_description_get)
+      expect(Project.find(project.id).line1).to eq('line1')
```
stuartmccoll: Can we assign the result of `Project.find(project.id)` to a variable and then assert against this instead of calling `.find` multiple times?

PR #356, opened 2020-03-19
Upload accounts page

paulmsmith:
Hi @stuartmccoll opening this PR for you to pick up.


app/jobs/application_to_salesforce_job.rb
```diff
@@ -37,6 +37,11 @@ def perform(project)
 
       ApplicationAttachmentsToSalesforceJob.perform_later(salesforce_case_id, project, :capital_work_file, "capital work attachment") if project.capital_work_file.attached?
       ApplicationAttachmentsToSalesforceJob.perform_later(salesforce_case_id, project, :governing_document_file, "governing document attachment") if project.governing_document_file.attached?
+
+      project.accounts_files.each_with_index do |af, idx|
+        ApplicationAttachmentsToSalesforceJob.perform_later(salesforce_case_id, project, :accounts_files, "accounts attachment ##{idx}", idx)
+      end
```
gidsg: If index is required to reference multiple attachments would that imply that `evidence_of_support` below is not working correctly and needs to use the same pattern? 
(Am I missing something?)

stuartmccoll: It's working differently here in that the model has a `has_many_attached :accounts_files` attribute which references a collection of `ActiveStorage` objects, whereas it has a `has_many :evidence_of_support` attribute which references nested objects. The way it's doing it for `evidence_of_support` worked fine for a nested object, but didn't work for a collection of attachments, hence the slight difference in the way we're passing it through for `accounts_files`.

gidsg: ahh, OK, makes sense.

app/javascript/error-handling/index.js
```diff
@@ -125,6 +125,8 @@ function createAndAppendListItem(message, attribute, modelName) {
 
         if (attribute === "governing_document_file") {
             linkElement.setAttribute("href", "#project_governing_document_file")
+        } else if (attribute === "accounts_files") {
```
gidsg: could we use a switch statement for all these conditions based on `attribute`?

PR #358, opened 2020-03-19
Adds target blank to each of the links on this page.

paulmsmith:
The work for ticket 1063


stuartmccoll:
Do these need '(link opens in new tab)' after each? I remember links opening in new tabs came up in conversation at some point, but can't remember what the outcome was!

paulmsmith:
@stuartmccoll I have not done that for now because I want to look at what the most accessible way is to do that. GOV.UK have guidance on it so for now I've not added it. I'll add a ticket to the backlog to define the pattern for external links.

PR #360, opened 2020-03-23
Updates reset password with govuk markup to make it look styled

paulmsmith:
The work to complete this ticket: https://trello.com/c/0SmwpZM0


app/views/devise/passwords/edit.html.erb
```diff
@@ -6,28 +6,29 @@
          }
 %>
 
-<h2>Change your password</h2>
+<h2 class="govuk-heading-l">Change your password</h2>
```
stuartmccoll:
```suggestion
<h1 class="govuk-heading-l">Change your password</h1>
```

Probably just a hangover from the original code, but guessing this should actually be a `h1`?

app/views/devise/passwords/edit.html.erb
```diff
@@ -6,28 +6,29 @@
...
-  <div class="actions">
-    <%= f.submit "Change my password" %>
+  <div class="govuk-form-group">
+    <%= render(ButtonComponent.new(element: "input", text: "Change my password")) %>
   </div>
-<% end %>
 
-<%= render "devise/shared/links" %>
+  <%- if devise_mapping.confirmable? && controller_name != 'confirmations' %>
+    <%= link_to "Didn't receive confirmation instructions?", new_confirmation_path(resource_name) %><br />
```
gidsg: what is this `<br />` doing. Could we remove it?

app/views/devise/passwords/edit.html.erb
```diff
@@ -6,28 +6,29 @@
...
 
-  <div class="field">
-    <%= f.label :password, "New password" %><br />
-    <% if @minimum_password_length %>
-      <em>(<%= @minimum_password_length %> characters minimum)</em><br />
-    <% end %>
-    <%= f.password_field :password, autofocus: true, autocomplete: "new-password" %>
+  <div class="govuk-form-group">
+    <%= f.label :password, "New password", class: "govuk-label" %>
+    <div class="govuk-hint"><% if @minimum_password_length %>Must be at least <%= @minimum_password_length %> characters<% end %></div>
```
stuartmccoll: Does it make more sense here to wrap the `div` inside the `if` condition? If `@minimum_password_length` was somehow not set, we'd end up rendering an empty `div`.

```suggestion
    <% if @minimum_password_length %><div class="govuk-hint">Must be at least <%= @minimum_password_length %> characters</div><% end %>
```

app/views/devise/passwords/edit.html.erb
paulmsmith: Good spot. Done.

app/views/devise/passwords/edit.html.erb
stuartmccoll: I think you resolved this but didn't commit the suggestion - unresolving it, but shout at me if that was intentional 😂!

PR #362, opened 2020-03-25
Added fix for IE11 AJAX handling

stuartmccoll:


app/views/partials/_direct_file_upload_hooks.js.erb
```diff
@@ -1,5 +1,33 @@
 "use strict";
 
+// TODO: Revisit this file to determine whether all aspects of it are
+// adding value or whether we can reduce this down.
```
gidsg: Would be good to TODO can this be moved into the main app JS?
(This will make it benefit from transpilation etc)

PR #364, opened 2020-03-31
Refactoring of address-related components

### LINKED ISSUE #72: Refactor postcode/address lookup functionality
Both the organisation and project sections of the service journey require the ability to lookup an address by postcode and to use the returned address to pre-fill form input values. At the moment, the code to do this is completely duplicated between the organisation and project controllers and views. We should refactor this to abstract it into a single place to make it easier to maintain.

stuartmccoll:
This pull request encompasses the work to refactor the address-related components within the application.

The address-related code has been moved into a single, reusable place, which is called from the other sections of the application using the following URL pattern:

`/<model_type>/<id>/rest-of-url`

We have also added validation of postcodes, to ensure that we're not calling the Ideal Postcodes API unnecessarily with data that we already know is invalid. As well, we've added caching of the address data that is returned from the Ideal Postcodes API, which means we no longer need to call the `/address` endpoint of the API with the UDPRN to retrieve the address details back. We've left this functionality in in the event that the cache is empty.

Apologies for the massive pull request, @gidsg! We took out some of the auto-generated boilerplate that was still hanging around at the same time, which has made this look bigger than it is!

Note that new tests have not yet been written, but existing ones have been modified to ensure that the pipeline will pass.

Closes #72


app/controllers/address_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+  # If correct, then assign the model type to a @type instance variable.
+  def check_and_set_model_type
+    if %w(user organisation project).include? params[:type]
+      @type = params[:type]
+      if @type == "organisation"
+        @model_object = Organisation.find(params[:id])
+      elsif @type == "project"
+        @model_object = Project.find(params[:id])
+      elsif @type == "user"
+        @model_object = User.find_by(organisation_id: params[:id])
```
gidsg: can we use a case statement here?

app/controllers/concerns/postcode_lookup.rb
```diff
@@ -1,124 +1,120 @@
...
-                                 postcode: @address_response.fetch(:postcode),
-                                 townCity: @address_response.fetch(:post_town),
-                                 county: @address_response.fetch(:county)
-                             })
+    # If cache has expired, retrieve from the API
+    unless @address_response.present?
+      logger.info "Address not found in cache, calling API using UDPRN: " \
+                  "#{params[:address]}"
+      lookup_address
+    end
```
gidsg: Can we express this more simply as:
```
    @address_response = Rails.cache.fetch("#{params[:address]}") do
      lookup_address
    end
```
as it accepts a block which will be used as a fallback see: https://guides.rubyonrails.org/caching_with_rails.html#low-level-caching

gidsg:
On production we should use the Redis cache store https://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-rediscachestore with the Cloudfoundry redis service. Let me know if you need a hand working out the config to get the credentials out of cloudfoundry.

app/helpers/postcode_helper.rb
```diff
@@ -1,5 +1,32 @@
...
+    [
+        [
+            [
+                result[:line_1], result[:line_2], result[:line_3]
+            ].reject(&:empty?)
+        ].join(', '),
+        result[:udprn]
+    ]
+
+  end
```
gidsg: Quite possibly missing something here but... not sure this is doing as the comment suggests? I think this would do, fetch this key from the cache if available or run the stuff in the `do` block? 
I think we should be setting the cache earlier on the API call when we assign `@response_array`? Which is passed to this function, which can then go back to purely doing the formatting as before?

app/views/address/postcode_lookup.html.erb
```diff
@@ -0,0 +1,55 @@
...
+} if flash[:errors] %>
+
+<%# todo: Move this to a _partial when headers are consistent across the service %>
+<% if @type == "organisation" || @type == "project" %>
+  <h1 class="govuk-heading-l" aria-label="Heading">About your <%= @type %></h1>
+<% else %>
+  <h1 class="govuk-heading-xl">
+    <span class="govuk-caption-xl">Add your details</span>
+    Find your address
+  </h1>
```
gidsg: seems a weird design thing to use different header sizes on these pages. Pinging @paulmsmith to review when he has a chance!

config/routes.rb
```diff
@@ -36,30 +42,32 @@
...
+        to: 'address#show'
+    put '/address-details',
+        to: 'address#update'
+    # This route ensures that attempting to navigate back to the list of address results
+    # redirects the user back to the search page
+    get '/address-results',
+        to: 'address#postcode_lookup'
+    # This route ensures that users can navigate back to the address details page
+    get '/address-attributes',
+        to: 'address#show'
```
gidsg: `attributes` seems quite technical language to expose in the URL. Could we call it `address-details` or something?

gidsg:
let me know if it's easier to discuss any of these comments in a call

app/views/address/postcode_lookup.html.erb
stuartmccoll: Agreed - we've essentially kept these as they were initially. I believe there's a card on the backlog somewhere to address the consistency between the headings.

app/helpers/postcode_helper.rb
stuartmccoll: Weirdly, you use `cache.fetch` to both read and write. if you call it with a `do` block, then it writes to the cache what you put inside the block. Doing it here means we can add only the bits of data we need beneath a unique key (the UDPRN), rather than stuffing the whole API response in the cache.

stuartmccoll: You're right about the function itself though - we've changed what it's doing without changing the name, so we can pick that up.

gidsg: yeah, if you need to keep this style. I would do the formatting part in a private method separately.

app/controllers/address_controller.rb
ptrelease: Yes we'll look at this.

app/controllers/concerns/postcode_lookup.rb
ptrelease: Thanks.  Will look into this.

app/helpers/postcode_helper.rb
ptrelease: We will change that.

config/routes.rb
ptrelease: Agreed

app/views/address/postcode_lookup.html.erb
stuartmccoll: I couldn't spot the original Trello ticket (if there was one!), so have created one [here](https://trello.com/c/cAlBvgT1).

manifest-production.yml
```diff
@@ -9,6 +9,7 @@ applications:
       - funding-frontend
       - log-drain
       - funding-frontend-uploads
+      - redis
```
gidsg, line 12: can we give this a bit more specific name like `funding-frontend-redis-cache`? This is because we might use redis instances for other distinct apps or purposes. (You can rename the service I think using the CF CLI)

app/controllers/concerns/postcode_lookup.rb
```diff
@@ -1,124 +1,141 @@
...
-          }
-        end
 
-      end
+  # Looks up a postcode from the Ideal Postcodes API. If the api returns an error, then
+  # Log the HTTP response code and API error code.  Then show the user a 500 page.
+  def lookup_postcode
+    begin
+      @response_array = IdealPostcodes::Postcode.lookup params['postcode']['lookup']
+      cache_address_results
```
gidsg, line 88: Very minor but could we pass `@response_array` in as a method argument? And perhaps similarly for `lookup_address` pass the param in as an argument. This I think should make the control flow a bit easier to understand and the methods easier to test (dependency injection style).

ptrelease: Agreed.

manifest-production.yml
ptrelease: Makes sense.

app/controllers/concerns/postcode_lookup.rb
```diff
@@ -1,124 +1,159 @@
...
+  # @return [array] response_array: An unformatted array of address results
+  def lookup_postcode(postcode)
+    begin
+      response_array = IdealPostcodes::Postcode.lookup postcode
+      cache_address_results(response_array)
+      response_array
+    rescue IdealPostcodes::IdealPostcodesError => e
+      logger.error "Error in Ideal Postcodes API.  Response body was: #{e.http_body} " \
+        "and HTTP code was:  #{e.http_code} and error code was: #{e.response_code}"
+      raise
```
gidsg, line 106: Can we pass a string to the exception, this is so if this happens the message with be logged e.g. in Sentry.
We could optionally create a custom exception class for this like we do e.g. in https://github.com/heritagefund/funding-frontend/blob/master/app/jobs/application_to_salesforce_job.rb#L4

The other option we have would be to flash an error message to the user e.g. _something went wrong, please try again if the problem persists contact us_ and log an error in Sentry in that case, rather than throw a 500 to the user. I don't mind too much raising though as I assume it will happen rarely.

PR #366, opened 2020-03-31
Feature/frontend changes

ptrelease:
Flipper changes.  Happy to discuss.


README.md
```diff
@@ -118,4 +118,22 @@ name you gave to the controller in the first command entered
...
+
+```postgresql
+insert into flipper_gates (feature_key, key, value, created_at, updated_at) Values('bau', 'boolean', true, NOW(), NOW());
+```
+
+Update the flipper row with:
+
+```postgresql
+update flipper_gates set value = false;
+```
```
gidsg: we can look at adding Flipper UI with auth to create this: https://github.com/jnunemaker/flipper/blob/master/docs/ui/README.md
but for now please can we do this as a rails database migration, rather than manually?

stuartmccoll: We had a think about this - we're initiating the app with the BAU flag disabled (Flipper does this by _not_ creating the database row). To enable BAU, we'd need to manually add the row. If we did add a row initially using a migration (with the flag set to `false`), we'd still need to manually update the row when we wanted to enable this. It felt like we weren't really getting any benefit from having the database migration other than running a different SQL statement to switch the flag (an `UPDATE` as opposed to an `INSERT`). Happy to chat it through, though 🙂.

gidsg:
LGTM, happy for you to Squash and merge on github as commit history is quite messy.

app/views/project/declaration/show_confirm_declaration.html.erb
```diff
@@ -101,10 +114,14 @@
                 {
                     attribute: "aria-label",
                     value: "Submit application button"
+                },
+                {
+                    attribute: ("disabled" if !Flipper.enabled?(:bau)),
+                    value: ("disabled" if !Flipper.enabled?(:bau))
```
gidsg, line 120:
```suggestion
                    attribute: ("disabled" unless Flipper.enabled?(:bau)),
                    value: ("disabled" unless Flipper.enabled?(:bau))
```

PR #368, opened 2020-04-03
Added validation for negative integers

stuartmccoll:
This commit adds validation for fields in the application which capture numerical values. These occur on the following pages - volunteers, non-cash contributions, cash contributions, and project costs. Previously, where no validation occurred, a user could enter a negative integer value and the application would pass this through. This could result in negative values being passed through to Salesforce, or a negative value being displayed on the overall grant request page.

Additionally, I have replaced the remaining numerical fields on the related views with text fields. Originally, we chose to use numerical fields to enable browser-based validation. However, GDS advise not to use input fields with a type attribute of number and to use input fields with a type attribute of text and to validate the value entered. We saw in user testing that the input fields with a type attribute of number do not work well with some assistive technologies.

### DIFF
--- a/app/models/cash_contribution.rb
+++ b/app/models/cash_contribution.rb
@@ -6,7 +6,10 @@ class CashContribution < ApplicationRecord
   has_one_attached :cash_contribution_evidence_files
 
   validates :description, presence: true
-  validates :amount, numericality: { only_integer: true }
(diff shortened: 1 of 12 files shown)


ptrelease:
Discussed with Stu and he's going to add some tests for this.

PR #370, opened 2020-04-06
move release form functionality to active job

gidsg:
This changes the existing webhook consumer functionality to an active job. This is so we can benefit from failure replay etc.


spec/controllers/released_form_controller_spec.rb
```diff
@@ -28,36 +30,7 @@
...
-          ApplicationStages: "Finance Approval",
-          ApplicationId: project.id,
-          AccountName: "New web test org",
-          AccountExternalID: "5188ff34-9011-4458-a9ea-19fe1ac9f3c6",
-          released_form: {},
-          ApprovedbyFinance:true}
-
-      expect(project.released_forms.first.form_type)
-          .to eq("completion_report")
+      } }.to have_enqueued_job
```
stuartmccoll, line 33: Could we be more specific about the job being enqueued here?

```suggestion
      } }.to have_enqueued_job(ReleaseFormJob)
```

app/controllers/released_form_controller.rb
```diff
@@ -2,13 +2,8 @@ class ReleasedFormController < ApplicationController
   http_basic_authenticate_with name: Rails.configuration.x.consumer.username, password: Rails.configuration.x.consumer.password
 
     protect_from_forgery with: :null_session
+
     def receive
```
stuartmccoll, line 6: I think that it'd be good to include a comment documenting the method - it's clear to see what it does, but it'd be good to record here how it's triggered.

PR #373, opened 2020-04-07
Refactored routes file and associated references

stuartmccoll:
We had a number of different sections in this file, some mapped to different models, such as `organisation` and `project`, and others for static pages.

We haven't been consistent with our approach in how we break out some of these routes into namespaces, which has left the file quite cluttered and created some debt.

The debt mostly exists where we haven't been consistent with our namespacing approach, which has led to us having to alias routes to then use them in controllers, tests, etc.

This pull request addresses this debt.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,5 +1,7 @@
 Rails.application.routes.draw do
 
+  # Devise root scope - used to determine the authenticated
+  # and unauthenticated root pages
   devise_scope :user do
     unauthenticated do
       root to: "devise/sessions#new"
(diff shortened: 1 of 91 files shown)


config/routes.rb
```diff
@@ -54,195 +81,109 @@
...
+        get 'cash-contributions', to: 'cash_contributions#show'
+        put 'cash-contributions', to: 'cash_contributions#update'
+        delete 'cash-contributions/:cash_contribution_id',
+               to: 'cash_contributions#delete',
+               as: :cash_contribution_delete
+        get 'your-grant-request', to: 'grant_request#show'
+        get 'are-you-getting-non-cash-contributions',
+            to: 'non_cash_contributions#question'
+        put 'are-you-getting-non-cash-contributions',
+            to: 'non_cash_contributions#question_update'
```
gidsg, line 133: I noticed that some of these methods use `#question` and `#question_update` rather than `#show` and `#update` we use in other places.
Is there a reason why or could we make these consistent with the other method names. (No problem if this is something better looked at in a separate PR)

stuartmccoll: The related methods exist in the same controller as the general cash/non-cash contributions pages. We could split them out separately in another pull request (would involve some additional refactoring - movement of views, tests).

PR #378, opened 2020-04-14
use sum rails helper instead of reduce

gidsg:
## Proposed changes
Small refactor. Use [sum](https://api.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-sum) Rails helper instead of reduce.


gidsg:
Note, we could use simpler syntax `object.sum(:amount)` but I think there were some existing nil amounts in the staging DB which would break.

Issue #124, opened 2020-01-13
display error if Direct Upload XHR request returns non 2xx response code


gidsg:
# Steps to reproduce
* Download this file, unzip it 
[payload.json.zip](https://github.com/heritagefund/funding-frontend/files/4053116/payload.json.zip)
* Upload the .json file on one of the file upload pages
# Actual Results
Silently fails 
<img width="982" alt="Screenshot 2020-01-13 at 09 52 20" src="https://user-images.githubusercontent.com/1764158/72249178-ba2a9f00-35f0-11ea-9403-376d0d24b9da.png">
# Expected Results
We should detect this state and display a user facing error message

gidsg:
@stuartmccoll I guess this forms part if we are doing a wider refactor to standardise file upload JS.

gidsg:
@stuartmccoll do you know if this is still an issue?

stuartmccoll:
I didn't refactor any of the things that Neil had done as part of the AJAX refactoring - was going to tackle that separately. If Neil's work didn't address this (I don't think it did), then I imagine it'll definitely still be a thing.

PR #381, opened 2020-04-15
bug fix: add id expected by direct upload JS

gidsg:
## Proposed changes
Previously in Sentry we have seen the error:
`Unable to get property 'prepend' of undefined or null reference`
I was able to reproduce this error by forcing the JS upload to fail by putting bad S3 config locally.
It appears the root of this error is:
https://github.com/heritagefund/funding-frontend/blob/4a6617f7c490eef599b18c8ef8755a1066607bcb/app/views/partials/_direct_file_upload_hooks.js.erb#L74
This is looking for an element in the DOM:
https://github.com/heritagefund/funding-frontend/blob/4a6617f7c490eef599b18c8ef8755a1066607bcb/app/views/partials/_direct_file_upload_hooks.js.erb#L57
...which did not exist on any of the file upload pages.

Adding this element resolves the issue and displays the contextual error message which is the intent of this code.
Screenshot 2020-04-15 at 16 38 45


app/views/project/accounts/show.html.erb
```diff
@@ -64,7 +64,7 @@
       <div id="accounts-form-group-main"
            class="govuk-form-group nlhf-highlight-content govuk-!-margin-bottom-6">
 
-        <div class="govuk-form-group" id="file-upload-group">
+        <div class="govuk-form-group" id="file-upload-form-group">
```
gidsg, line 67: I noticed this page and _governing documents_ used an id `file-upload-group` whereas the others used ids like `capital-works.capital_work_files-form-group` do you know the rationale behind this @stuartmccoll?

gidsg:
Also, that means I think we have at least 3 different ways of displaying errors on these pages 🤯 
Probably ripe for a review or refactor?

app/views/project/accounts/show.html.erb
stuartmccoll: I _think_ it stems from a mixture of the original error handling written by Neil, and then the updated stuff which I refactored. That's a guess at the moment, though. I'd like to pull this branch down and run some manual tests before I put a 👍 on it - will get to that at some point this afternoon.

gidsg:
@stuartmccoll 
> I was able to reproduce this error by forcing the JS upload to fail by putting bad S3 config locally.

If you want to try reproducing this, I put this in my local `storage.yml` to make it fail the upload to S3:
```
local:
  service: S3
  access_key_id: <%= Rails.env.staging? ? ENV.fetch('AWS_ACCESS_KEY')  : "" %>
  secret_access_key: <%= Rails.env.staging? ? ENV.fetch('AWS_SECRET_KEY') : "" %>
  region: <%= Rails.env.staging? ? CF::App::Credentials.find_by_service_label('aws-s3-bucket')['aws_region'] : "eu-west-2" %>
  bucket: <%= Rails.env.staging? ? CF::App::Credentials.find_by_service_label('aws-s3-bucket')['bucket_name'] : "eu-west-2" %>
```

gidsg:
> I think this is possibly down to the governing documents and accounts pages not having the file upload instance variable set within their show methods 

@stuartmccoll yes, this was the issue, I have pushed a fix now.

gidsg:
fixes #124

PR #383, opened 2020-04-21
Automated tests for support pages

ptrelease:
## Proposed changes

Trello card #1385.  Added unit and integration tests for the support controller.

## Further comments

Not large or complex.  However, we may consider refactoring to validate in a support model at a later stage.  Also we made the clear_flash method public in order to test - which isn't ideal - and could be removed as part of the later refactor.

## Screenshot (if applicable)
N/A
## Checklist


app/views/support/report_a_problem.html.erb
```diff
@@ -38,13 +38,13 @@
       </div>
 
       <div class="govuk-form-group <%= "govuk-form-group--error" if
-                                           flash[:errors] && flash[:errors][:support_problem_message] %>">
+                                           flash[:errors].present? && flash[:errors][:support_problem_message].present? %>">
```
gidsg, line 41: Rather than two checks could we do a safe lookup inside the hash e.g. using Rails [try](https://api.rubyonrails.org/classes/Object.html#method-i-try) helper?

PR #385, opened 2020-04-21
Revert "Send Devise notifications asynchronously"

gidsg:
Revert this to allow time to investigate broken links: https://trello.com/c/LgVOvR5q/1431-devise-links-in-notify-emails-are-broken
Reverts heritagefund/funding-frontend#379

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -50,14 +50,7 @@ class User < ApplicationRecord
     end
   end
 
-  def self.current_user(user_id)
-      @user = User.find_by(uid: user_id)
-  end
-
-  # Override the send_devise_notification method to allow for
-  # non-blocking sending using ActiveJob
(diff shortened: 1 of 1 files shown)


stuartmccoll:
Happy to revert if you think it's necessary - testing both `master` and this branch locally gives me working URLs in the Devise emails I'm getting, which makes me think it's something configuration related, as you mentioned previously.

gidsg:
@stuartmccoll this works differently on production see: https://github.com/heritagefund/funding-frontend/blob/75c865bb9e3bb4d0b4829c2be155ef5c028318ab/config/environments/production.rb#L128

gidsg:
this value is resolving fine on the box if I SSH in but maybe something different about how the worker reads configuration or resolves its environment?

PR #389, opened 2020-04-23
add Open Graph metadata

gidsg:
## Proposed changes

Add open graph metadata for previews on Facebook, Slack etc.

@paulmsmith note, currently I am just re-using the apple touch image. We may want to supply a high res image as described in: https://developers.facebook.com/docs/sharing/best-practices#images


gidsg:
Would rather get a better image asset from @paulmsmith before merging if possible. Else we can leave as a TODO.

paulmsmith:
@gidsg I will add the image, sorry for the delay been on other website tasks and a lot of calls. :)

paulmsmith:
Added open graph image. Over to you @gidsg 👍

PR #391, opened 2020-04-24
actionview-component has been renamed to view_component

gidsg:
## Proposed changes

Updates `actionview-component` as it has been renamed to `view_component`, see: https://github.com/github/view_component#migration-from-actionviewcomponent 

## Checklist


config/environments/development.rb
```diff
@@ -77,8 +77,6 @@
 
   config.x.support_email_address = 'test@example.com'
 
-  config.action_view_component.preview_path =
-      "#{Rails.root}/spec/components/previews"
```
stuartmccoll, line 81: @paulmsmith was using this to test and create components - I'm guessing that by removing the configuration, he'd no longer be able to do that.

Assuming the configuration still exists, I imagine it'll change to `config.view_component.preview_path` the same as other settings have.

PR #393, opened 2020-04-24
fix duplicate ID and missing ARIA label

gidsg:
## Proposed changes

fixes issues found in accessibility review https://trello.com/c/IOxJbSsw

### DIFF
--- a/app/views/project/dates/show.html.erb
+++ b/app/views/project/dates/show.html.erb
@@ -161,7 +161,7 @@
           Project end date
         </legend>
 
-        <span id="project-start-hint" class="govuk-hint">
+        <span id="project-end-hint" class="govuk-hint">
           For example, 29 4 2020
         </span>
 


app/views/project/dates/show.html.erb
```diff
@@ -161,7 +161,7 @@
           Project end date
         </legend>
 
-        <span id="project-start-hint" class="govuk-hint">
+        <span id="project-end-hint" class="govuk-hint">
```
stuartmccoll, line 164: Looks fine in terms of fixing the duplicate `id` attribute, but I can't see the addition of an `aria-label`?

gidsg: the aria label was failing to work previously as it was expecting the ID to match that defined here: https://github.com/heritagefund/funding-frontend/blob/e2473a2dbc5b32f6ad2761a2622a86328f7491e8/app/views/project/dates/show.html.erb#L157

stuartmccoll: 👍

PR #409, opened 2020-06-24
Replicate user details to associated person

stuartmccoll:
## Proposed changes

This pull request adds functionality to replicate a user's details to an associated person record.

A `person` is created at the point a user registers for this service, with the user's email being replicated to the `person` record. When a user signs in and adds further details - `name`, `phone_number` and `date_of_birth` - these are all replicated to the associated `person`. When a user then adds their `address`, a new `address` record is created which is associated to the `person` via a record on the `persons_addresses` table.

Existing functionality which stores details on the `users` record remains unchanged.

## Further comments

Further consideration will need to be given to a `user` updating their address at a later point. Should the associated address be overwritten, or should a new address be created and the association updated to reflect this? If we later choose to only store one record for an address - e.g. not duplicating records where addresses are identical - then we should choose the second option previously mentioned.

Any breaking tests have been fixed, but no new tests have been added.

Further, the `users_addresses` migration has been removed, as this is superfluous.

## Checklist


app/controllers/address_controller.rb
```diff
@@ -70,6 +75,71 @@ def check_and_set_model_type
...
+  # Checks for the presence of an associated Address via the PeopleAddresses
+  # model. If none exists, a new Address and PeopleAddress is created. Address
+  # details are replicated from the current_user object
+  #
+  # @param [User] user An instance of User
+  def check_and_set_person_address(user)
+
+    person_address_association = PeopleAddress.find_by(person_id: user.person_id)
+
+    unless person_address_association
```
ptrelease, line 87: consider found_person_address_association?

Reads better in subsequent code?

PR #416, opened 2020-07-22
Feature/alternative finance form design

paulmsmith:
## Proposed changes

- Adds some space and minor design tweaks to hopefully make the form a little more readable.
- Adds the 'for' attribute to checkbox labels to improve accessibility


app/views/funding_application/hef_loan/form/show.html.erb
```diff
@@ -1,3 +1,10 @@
+<%=
+  render partial: "partials/page_title",
+         locals: {
+             model_object: nil,
```
stuartmccoll, line 4:
```suggestion
             model_object: @funding_application
```

PR #422, opened 2020-07-27
added validation for loan amount

ptrelease:
## Proposed changes

At the moment, loan amount isn't checked for an appropriate value. This change fixes that.
Tests not explicitly coded.  This will follow in a later debt card.


config/locales/en.yml
```diff
@@ -121,7 +121,7 @@ en-GB:
               blank: "Enter details about how you plan to spend the money"
             loan_amount_requested:
               not_a_number: "Loan amount being requested must be a number, like 500"
-              greater_than: "Enter an amount greater than 0"
```
stuartmccoll, line 124: Is the new `inclusion` error message displayed if an amount less than `0` is entered? If so, it's safe to remove the `greater_than: 0` part of the validation. If now, we should leave the `greater_than` error message in.

ptrelease: Yes.  Same error message is displayed.

PR #465, opened 2020-08-21
Linted controllers/concerns files

stuartmccoll:
## Proposed changes

This commit contains fixes to reported RuboCop violations in files within the `app/controllers/concerns` directory.

## Further comments

N/A.

## Screenshot (if applicable)

N/A.

## Checklist


.rubocop.yml
```diff
@@ -2,15 +2,23 @@ Layout/EmptyLinesAroundClassBody:
...
 Metrics/MethodLength:
   Severity: warning
+Metrics/ModuleLength:
+  Severity: warning
 Metrics/PerceivedComplexity:
   Severity: warning
 
 Style/FrozenStringLiteralComment:
   Enabled: false
+Style/RescueStandardError:
```
ptrelease, line 23: Looks like we don't infringe this often - and its a fair cop.  Would it be better to make the few we have exceptions - so we can keep this cop?

ptrelease: Absolutely fine otherwise

ptrelease: Waiting for rubocop docs to come back up to check this.

ptrelease: Explicit fine.

ptrelease (approved):

PR #471, opened 2020-08-28
Linting changes made to all controllers in the organisation folder

ptrelease:
Linting changes made.

Rubocop only reports warnings now.  And RSpec runs without errors. Manaully tested the organisation pages and found no errors.


app/controllers/organisation/numbers_controller.rb
```diff
@@ -18,8 +20,10 @@ def update
 
     else
 
+      # rubocop:disable Style/StringLiterals
       logger.info "Validation failed when attempting to update company_number/charity_number " \
                     "for organisation ID: #{@organisation.id}"
+      # rubocop:enable Style/StringLiterals
```
ptrelease, line 27: I could have broken the string up.  But I figured one us would try and put it back to a single statement - then encounter this again.

app/controllers/organisation/mission_controller.rb
```diff
@@ -33,19 +33,18 @@ def update
...
 
     # When no checkbox is selected on the page no mission key/value is passed
     # in the form, meaning that the organisation hash is no longer passed
     # through either. In this case, we need to add it manually to avoid
     # triggering a 'param is missing or value is empty' exception
-    unless params[:organisation].present?
-      params.merge!({organisation: {mission: ""}})
-    end
+
+    params.merge!({ organisation: { mission: '' } }) if params[:organisation].blank?
```
stuartmccoll, line 44:
```suggestion
    params.merge!({ organisation: { mission: '' } }) unless params[:organisation].present?
```

I'm not convinced by the change in logic here - we've kept to `.present?` elsewhere when addressing RuboCop violations.

I'll test locally and then commit the suggestion if it doesn't re-trigger a violation.

app/controllers/organisation/numbers_controller.rb
stuartmccoll: I think we should be fine to just put the first line in `'`'s and then the second line in `"`'s. I'll test locally and remove the RuboCop comments if so.

stuartmccoll (changes requested):

app/controllers/organisation/numbers_controller.rb
stuartmccoll: Tested locally and this works as expected.

stuartmccoll (approved):

PR #472, opened 2020-08-28
linted controllers in the user folder

ptrelease:
Linting changes made.

Rubocop only reports warnings now. And RSpec runs without errors. Manually tested the user and registration pages and found no errors.


stuartmccoll (approved):
Looks fine, but I'll give the `check_and_set_organisation` another manual test locally to be on the safe side.

PR #482, opened 2020-11-17
Abstract the link between users and organisations

stuartmccoll:
## Proposed changes

This pull request contains commits which update funding-frontend to abstract the link between `users` and `organisations`. A user should be able to apply on behalf of more than one organisation, and as such this association has been abstracted into a `users_organisations` association.

References to a user's organisation now point to `users.organisations.first`, where this will change in the future when functionality in this area is introduced.

## Further comments

N/A.

## Screenshot (if applicable)

N/A.

## Checklist

- [X] Updated `README.md` if necessary


db/migrate/20201113090359_create_users_organisations.rb
```diff
@@ -0,0 +1,10 @@
+class CreateUsersOrganisations < ActiveRecord::Migration[6.0]
+  def change
+    create_table :users_organisations, id: :uuid do |t|
+      t.references :user, type: :int, null: false, foreign_key: true
+      t.references :organisation, type: :uuid, null: false, foreign_key: true
+
+      t.timestamps
+    end
+  end
+end
```
ptrelease, line 10: Are we stuck with integer as the id for a user?

db/migrate/20201118085253_remove_organisation_from_users.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveOrganisationFromUsers < ActiveRecord::Migration[6.0]
+  def change
+    remove_column :users, :organisation_id, :uuid
+  end
+end
```
ptrelease, line 5: Have we confirmation that orgs affiliates to a user are not needed?

db/migrate/20201113090359_create_users_organisations.rb
stuartmccoll: Yep - Devise.

db/migrate/20201118085253_remove_organisation_from_users.rb
stuartmccoll: 🤔 They're still affiliated, through the associative table now.

ptrelease (approved):

PR #509, opened 2021-01-05
Setting up standardrb and rubocop

naps62:
## Proposed changes

This sets up two linters:
- [`standard`](https://github.com/testdouble/standard), which is, as the name suggests, meant to be the standard styleguide for Ruby.
- rubocop, which was already set up, but is updated and re-configured here. standard relies on rubocop underneath, but doesn't implement everything (it's still in beta, and a work in progress). I usually rely on both, and configure rubocop to explicitly enforce some very useful rules that editors need to take into account (such as how do indent parameters, blocks, etc). So rubocop now is configured with `DisabledByDefault` to disable all rules, and then specifically enable only the ones that standard doesn't take care of

99.9% of the file changes here were done by running `standardrb --fix` and `rubocop -A`. Only a handful of changes needed manual editing

It should be noted that any configuration here is up for debate. If someone disagrees with a particular lint, that can always be discussed and changed

## Further comments

This isn't yet set up on the CI, although that's a very simple addition. We'd only need to add `standardrb` and `rubocop` as part of the github actions workflow.
But I'd prefer to wait a bit on that to see how the rest of the team feels about these changes, before enforcing rules that might break/delay existing work. Once everyone understands and is confortable with this, we can start enforcing them more strongly

VS Code support should be easy to get with the main ruby plugin: https://github.com/rubyide/vscode-ruby.git


app/models/application/common/funding_application.rb
```diff
@@ -18,68 +17,62 @@ class FundingApplication < ApplicationRecord
   attr_accessor :validate_declarations
 
   validates_associated :organisation
-  validates_associated :people if :validate_people
```
naps62, line 21: FYI: you may notice this one was just wrong, `if :validate_people` would run the condition at compille-time, and would always fail, meaning this validation would never even be applied to the model

PR #510, opened 2021-01-05
Run CI pipeline only on PRs instead of every commit

naps62:
When pushing commits to a work-in-progress branch, the current pipeline on Github Actions will be executed.

This can be a waste of resources (and potentially exceed the free-tier limits of Github Actions) as well as delay actually relevant jobs.

Instead, it's a better practice to just build on one of two conditions:
- any commit to the `master` branch
- after a Pull Request is opened or updated

This way, when opening a PR, the pipeline will be executed for it. But until the PR is opened, we won't be wasting resources.


stuartmccoll (approved):
Thanks @naps62, this looks good 👍.

PR #513, opened 2021-01-06
Add pre-application journeys with stubbed content

ptrelease:
Initial changes to introduce pre-applications such as project enquiries and expressions of interest to funding-frontend 

Added new models, views, controllers and migrations to accomplish this


app/controllers/pre_application/expression_of_interest/feasibility_or_options_work_controller.rb
```diff
@@ -0,0 +1,43 @@
+# Controller for the Pre-application 'Start' page
```
ptrelease, line 1: Change this

app/controllers/pre_application/expression_of_interest/likely_submission_description_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+    @pre_application.pa_expression_of_interest.update(pa_expression_of_interest_params)
+
+    if @pre_application.pa_expression_of_interest.valid?
+
+      logger.info 'Finished updating likely_submission_description for pa_expression_of_interest ID: ' \
+                  "#{@pre_application.pa_expression_of_interest.id}"
+      
+      the_json = @pre_application.to_salesforce_json
+
+      logger.info(the_json)
```
ptrelease, line 24: Not sure we should be logging,  consider how long logs held for.  May be better to store on DB, so we can easily clear later as batch or scheduled job.  Same applies to PEF.

app/controllers/pre_application/expression_of_interest/submitted_controller.rb
```diff
@@ -0,0 +1,5 @@
+# Controller for the Pre-application 'Start' page
+class PreApplication::ExpressionOfInterest::SubmittedController < ApplicationController
+    include PreApplicationContext
+
+end
```
ptrelease, line 5: add newline

ptrelease:
@stuartmccoll Looks good to me - Can't approve myself but happy when you are.

stuartmccoll (approved):

PR #515, opened 2021-01-06
Better handling of date of birth fields

naps62:
Instead of manually handling parsing and error checks for day/month/year fields, we can treat the full date as a single entity.

Both rails params and activemodel callbacks automatically handle multipart fields such as dates and times
Only a small quirk was needed, because in the event of an invalid date (e.g.: 32nd of January), instead of a model error, an exception is raised.

The solution is to patch `ActiveRecord::Base.update` to handle that exception, by adding the corresponding errors to the model and return gracefully

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-git_source(:github) { |repo| "https://github.com/#{repo}.git" }
+git_source(:github) do |repo| "https://github.com/#{repo}.git" end
 
 ruby '~> 2.6.0'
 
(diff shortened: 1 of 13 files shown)


stuartmccoll (commented):
I don't _think_ this is working quite as intended - at the minute, if I enter a date of birth of 31/2/1920 on the user details page, then the date of birth saved to the database is 3/2/1920. Likewise, for a data of birth of 31/11/2020 (a day which doesn't exist), the code seems to pass through an add a date of birth to the database of 1/12/2020. I'd expect to see an 'invalid date'-type error in these cases, which I think the current code is handling using `Date.valid_date?`.

naps62:
Hey @stuartmccoll 
just iterated on this

The comment when submitting an empty/incomplete date was an oversight on my part. I changed the code a bit before opening the PR, and somehow messed it up. It's good now

As for the 31/2/1920, it's something I just learned right now as well :smile: . Apparently the inner workings of Ruby mean that February 31st is parsed as March 2nd.
I added a validator gem that allows us to deal more strictly with those. I've used the gem before, and highly recommend it. I just didn't know it was necessary for this particular scenarios

Anyway, there's still a slight problem to solve, regarding error messages:

image


But that seems to also be an issue with the way your code displays those error messages, so I'll handle that as a separate PR

PR #518, opened 2021-01-08
Using strong_params with potentially empty params

### LINKED ISSUE #40: Improve form validation to combine Active Record and strong parameters validations
The validation checks used against forms within the service could be made more robust by combining the [validation checks provided by Active Record](https://guides.rubyonrails.org/active_record_validations.html) and by the use of [strong parameters](https://edgeapi.rubyonrails.org/classes/ActionController/StrongParameters.html).

We've struggled to get these to work in conjunction when a user has submitted an empty form (such as not selecting a radio button), where the strong parameters validation checks then fail because the model object has not been sent through in order to fulfil the `.require()` check. It seems like it _should_ be possible to get both checks to work together, but this will require some time invested into exploring this further.

naps62:
Fixes #40, although it should need a review to double-check I'm not missing any edge case.
The test suite for this controller seems complete enough, so no new tests were added

(all relevant discussion and explanation of the issue is already detailed on #40 )


stuartmccoll (approved):
Thanks @naps62, I've tested and this works as expected :+1.

PR #530, opened 2021-01-13
best_placed string abstraction complete

ptrelease:
Abstracted strings on best_placed view into a local file.


app/views/funding_application/gp_project/best_placed/show.html.erb
```diff
@@ -63,7 +62,6 @@
       <span id="project_best_placed_description-info"
             class="govuk-hint govuk-character-count__message"
             aria-live="polite">
-        You can enter up to 500 words
```
stuartmccoll, line 66: I think you should put the generic max words translation in here, as I think that kicks in if JS is turned off. It's something like `<%= t('generic.word_count', max_words: 500) %>`.

stuartmccoll (changes requested):

app/views/funding_application/gp_project/best_placed/show.html.erb
ptrelease: Ah, that makes sense

stuartmccoll (approved):

PR #538, opened 2021-02-03
Added SalesforceApi library

stuartmccoll:
## Proposed changes

This commit adds a `SalesforceApi` library, enabling creation of a `Restforce` client and retrieval of payment-related details for a funding application.

## Further comments

N/A.

## Screenshot (if applicable)

N/A.

## Checklist

- [X] Updated `README.md` if necessary


lib/apis/salesforce/salesforce_api.rb
```diff
@@ -0,0 +1,86 @@
...
+    #
+    # @return [Hash] A Hash, currently containing only the amount awarded to the project
+    #                and the percentage of the total costs that the organisation have agreed
+    #                to award
+    def get_payment_related_details(id)
+
+      Rails.logger.info("Retrieving payment-related details for project ID: #{id}")
+
+      begin
+
```
ptrelease, line 30: Does the below comment still apply?  Can't see grant % on line 31

lib/apis/salesforce/salesforce_api.rb
```diff
@@ -0,0 +1,86 @@
...
+        api_version: '47.0'
+      )
+
+      Rails.logger.info('Finished initialising Salesforce client')
+
+    end
+
+  end
+
+end
```
ptrelease, line 86: I like this.  I can see this module getting massive though, so at some point we may need to consider braking apart, or inheriting modules from another with the initialise_client method.

ptrelease: How are you planning to instantiate and use salesforce_api.rb itself? Not something we need to worry about too much right now - but this implementation suggests you will implement as a singleton (which is what I'm hoping)

lib/apis/salesforce/salesforce_api.rb
```diff
@@ -0,0 +1,86 @@
...
+        security_token: Rails.configuration.x.salesforce.security_token,
+        client_id: Rails.configuration.x.salesforce.client_id,
+        client_secret: Rails.configuration.x.salesforce.client_secret,
+        host: Rails.configuration.x.salesforce.host,
+        api_version: '47.0'
+      )
+
+      Rails.logger.info('Finished initialising Salesforce client')
+
+    end
```
ptrelease, line 82: Todo: look up whether we need to close this connection

lib/apis/salesforce/salesforce_api.rb
stuartmccoll: Fixed in subsequent commit.

lib/apis/salesforce/salesforce_api.rb
stuartmccoll: Discussed offline.

lib/apis/salesforce/salesforce_api.rb
stuartmccoll: Discussed offline.

ptrelease (approved):

PR #545, opened 2021-02-10
Abstract footer and content

paulmsmith:
## Proposed changes

The work is described in this Trello ticket: https://trello.com/c/rwCc0CEZ/1758-abstract-content-from-footer-in-funding-frontend


paulmsmith:
@stuartmccoll I wasn't sure if I should create the cy.yml file yet? I have done because that it what I'm guessing we will send for translation?

ptrelease (approved):
Code looks good, runs in server and tests pass.  We don't typically create the CY file, but no problem if that's how you'd like to work.

stuartmccoll (approved):

paulmsmith:
Thanks @ptrelease I just thought it needs creating so I could do it as part of this bit of work and then edit it with the translation when I get that back.

PR #555, opened 2021-02-13
Feature/2287 community importance

paulmsmith:
## Proposed changes

[Ticket is here](https://trello.com/c/MybwOuoY/2287-abstract-strings-from-project-importance-to-community-page)


app/views/funding_application/gp_project/matter/show.html.erb
```diff
@@ -27,17 +27,16 @@
                                          @funding_application.project.errors[:matter].any? %>">
 
       <h1 class="govuk-label-wrapper">
-        <span class="govuk-caption-xl">About your project</span>
+        <span class="govuk-caption-xl"><%= t('views.funding_application.common.about_your_project') %></span>
         <%=
           f.label :matter,
-                  "Why is your project important to your community?",
+                   t('application.matter.headings.1'),
```
stuartmccoll, line 33: Can we use `page_heading` here instead?

app/views/funding_application/gp_project/matter/show.html.erb
```diff
@@ -59,7 +58,7 @@
       <span id="project_matter-info"
             class="govuk-hint govuk-character-count__message"
             aria-live="polite">
-        You can enter up to 500 words
+        <%= t('application.matter.wordcount') %>
```
stuartmccoll, line 61:
```suggestion
        <%= t('generic.word_count', max_words: 500) %>
```

config/locales/funding_application/gp_project/matter/cy.yml
```diff
@@ -0,0 +1,9 @@
+cy:
+  application:
+    matter:
+      page_title: "Why is your project important to your community?"
+      headings:
+        1: "Why is your project important to your community?"
+      text:
+        1: "This helps us understand how engaged your community is with your project and why it is important to them."
+      wordcount: "You can enter up to 500 words"
```
stuartmccoll, line 9: Safe to remove this section once using the generic word count localisation.

config/locales/funding_application/gp_project/matter/en.yml
```diff
@@ -0,0 +1,9 @@
+en-GB:
+  application:
+    matter:
+      page_title: "Why is your project important to your community?"
+      headings:
+        1: "Why is your project important to your community?"
+      text:
+        1: "This helps us understand how engaged your community is with your project and why it is important to them."
+      wordcount: "You can enter up to 500 words"
```
stuartmccoll, line 9: Safe to remove this section once using the generic word count localisation.

stuartmccoll (changes requested):

app/views/funding_application/gp_project/matter/show.html.erb
paulmsmith: We could but I thought I might be future proofing incase one needed to be different from the other?

stuartmccoll: Sorry, I probably didn't explain that very well. We always keep `page_title` and `page_heading` separate so that they can differ, with the title being used as the value of the `title` element and the `page_heading` being used as the page's `h1` element. Does that make sense?

paulmsmith: Ah OK. That does make sense.

config/locales/funding_application/gp_project/matter/cy.yml
```diff
@@ -0,0 +1,8 @@
+cy:
+  application:
+    matter:
+      page_title: "Why is your project important to your community?"
+      page_heading: "Why is your project important to your community?"
+      text:
+        1: "This helps us understand how engaged your community is with your project and why it is important to them."
+      wordcount: "You can enter up to 500 words"
```
stuartmccoll, line 8: Remove `wordcount` key/val pairing from this file.

stuartmccoll (changes requested):
Have left a comment - just the `wordcount` key/val pair to remove from `cy.yml`.

ptrelease (approved):

ptrelease (approved):

PR #578, opened 2021-02-25
2303 -  adds welsh translation to yml files

paulmsmith:
## Proposed changes

This is the ticket to add into funding front-end the translation strings provided by Liam. See the [ticket](https://trello.com/c/rvgLAc5U/2302-update-the-abstracted-string-files-with-liams-translations-in-a-single-pull-request)


paulmsmith:
The review needed for this is a technical one. Check I didn't break the site and that Welsh does show on the pages affected. Liam is going to do a manual review on staging when this is merged.

config/locales/partials/footer/cy.yml
```diff
@@ -1,9 +1,9 @@
 cy:
   footer:
-    copyright: "Trustees of the National Heritage Memorial Fund"
+    copyright: "Ymddiriedolwyr Cronfa Goffa'r Dreftadaeth Genedlaethol"
     links:
-      english: "English"
+      english: "Saesneg"
```
ptrelease, line 5: This is actually a little confusing if you are an English speaker returning to Welsh.  Perhaps leave as English?

config/locales/support/question_or_feedback/cy.yml
```diff
@@ -1,14 +1,13 @@
...
-      your_email_address: "Your email address"
\ No newline at end of file
+      your_message_was_sent_successfully: "Anfonwyd eich neges yn llwyddiannus"
+      service_status:
+        html: "%{href} i weld a oes unrhyw faterion hysbys."
+        text: "Edrychwch ar ein tudalen statws gwasanaeth"
+    form:
+      your_name: "Eich enw"
+      your_message: "Eich neges"
+      your_email_address: "Eich cyfeiriad e-bost"
```
ptrelease, line 13: This page isn't toggling to Welsh.  I think you've brought the 'report a problem' content in here, so the translation structure is different and nothing is kicking in.

config/locales/cy.yml
```diff
@@ -1,7 +1,359 @@
...
+    evidence_attached: "Tystiolaeth ynghlwm"
+    click_to_view_file: "Cliciwch i weld y ffeil"
+    total: "Cyfanswm"
+    amount: "Swm"
+    description: "Disgrifiad"
+    for_example: "er enghraifft"
+  views:
+    funding_application:
+      common:
+        about_your_project: "Ynglŷn â'ch prosiect"
```
ptrelease, line 40: Can you check the translation for about you project kicks in on yours?  Doesn't on mine.  YML seems OK, so curious.

ptrelease: Ah - its en_GB at the top of the file.  Will change.

config/locales/partials/footer/cy.yml
paulmsmith: Hmmm. Yeh, I think you're right although I can see it the other way around too. Will change.

config/locales/cy.yml
```diff
@@ -1,7 +1,359 @@
...
+        support_for_your_project: "cefnogaeth i'ch prosiect"
+      gp_project:
```
ptrelease, line 42: Common has 4 things in the english yml vs 2 things in the welsh yml (2 are missing)

ptrelease: Also views.funding_application.common.support missing from bothe yml files (used in evidence of support)

paulmsmith: > Also views.funding_application.common.support missing from bothe yml files (used in evidence of support)

How has that happened? Merge issue?

config/locales/support/question_or_feedback/cy.yml
paulmsmith: Hmm. This is how it was supplied by our translator. I think something might have got messed up in sending a zip of yml files.

paulmsmith: Will remove this YML for now until we can get the correct translation.

config/locales/cy.yml
ptrelease: Will pick up in a later PR when we have the translations.

config/locales/support/question_or_feedback/cy.yml
ptrelease: Paul will remove this

ptrelease (approved):

PR #627, opened 2021-03-15
Added controller tests for EOI StartController

stuartmccoll:
## Proposed changes

This commit adds RSpec controller tests for the `PreApplication::ExpressionOfInterest::StartController`.

## Further comments

N/A.

## Screenshot (if applicable)

N/A.

## Checklist


spec/controllers/pre_application/expression_of_interest/start_controller_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+      subject.current_user.organisations.clear
+
+      put :update
+
+      expect(subject.current_user.organisations.count).to(eq(1))
+
+      # Expect that a PreApplication and an associated PaExpressionOfInterest
+      # have been created
+      expect(assigns(:pre_application)).not_to(be_nil)
+      expect(assigns(:pre_application).pa_expression_of_interest).not_to(be_nil)
```
ptrelease, line 21: Should we check the assigned user too?

ptrelease: (to pre_application)

ptrelease:
@stuartmccoll With more time, think we should have written three tests for this start_controller.  One to check that the update method calls orchestrate_expression_of_interest_start_journey with current_user.  A second to check that orchestrate_expression_of_interest_start_journey makes its inner scope calls with user and organisation.  And you've covered the third to get the pre-app creation. 

But these tests to implicitly covers the helper functions are great with the time we have.  Just not a pattern I'd like to move forward with.

ptrelease (approved):

PR #637, opened 2021-03-16
Abstracted funding application project title markup into partial

stuartmccoll:
## Proposed changes

This commit abstracts the funding application project title markup into a partial so that it can be referenced by multiple application journeys.

## Further comments

N/A.

## Screenshot (if applicable)

N/A.

## Checklist


stuartmccoll:
@ptrelease I've done this one as an example of how quick it'll be to partial-ise the current £3,000 to £10,000 funding application journey. Once the markup is sat in a partial, it's a tiny change to make the current £3,000 to £10,000 question utilise the partial.

I'm sure there'll be one or two that are more complex, but for the most part they're really quick to do.

ptrelease (approved):

PR #654, opened 2021-03-19
Abstracted funding application project volunteers into partial

stuartmccoll:
## Proposed changes

This commit abstracts the funding application project volunteers markup into a partial so that it can be referenced by multiple application journeys.

## Further comments

N/A.

## Screenshot (if applicable)

N/A.

## Checklist


app/views/funding_application/gp_project/volunteers/show.html.erb
```diff
@@ -19,7 +22,7 @@
       <ul class="govuk-list govuk-error-summary__list">
 
         <% @funding_application.project.errors.each do |attr, msg| %>
-          <% unless attr.to_s == "volunteers" %>
+          <% unless attr.to_s == 'volunteers' %>
```
ptrelease, line 25: Should we iterate over errors here - then check on errors.attribute?

ptrelease (approved):

PR #702, opened 2021-04-06
Added formal heritage designations question to medium grants

stuartmccoll:
## Proposed changes

This commit adds a new question to the open medium grant programme journey.

## Further comments

N/A.

## Screenshot (if applicable)

N/A.

## Checklist


ptrelease:
Bugs seen in testing:

- If you exceed the word limit for one text field, the link at the top of the page doesn't go there. If you have more than one error it's fine.
- If you exceed a word limit in a text field, then uncheck that field, validation still occurs against it preventing an applicant moving forward

app/controllers/funding_application/gp_open_medium/heritage_designations_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+      end
+  
+    end
+  
+    private
+  
+    def open_medium_params
+  
+      params.fetch(:open_medium, {}).permit(
+        :hd_grade_1_description,
```
ptrelease, line 47: Should be  :hd_grade_1_a_description?,

app/views/partials/funding_application/_heritage_designations_question.html.erb
```diff
@@ -0,0 +1,175 @@
+<div class="govuk-form-group <%= "govuk-form-group--error" if model_object.errors.any? %>">
+
+  <fieldset class="govuk-fieldset" aria-describedby="mission-hint">
```
ptrelease, line 3: consider changing the aria

app/views/partials/funding_application/_heritage_designations_question.html.erb
```diff
@@ -0,0 +1,175 @@
...
+          ) if hd.designation == 'other'
+        %>
+
+      <% end %>
+
+    </div>
+
+  </fieldset>
+
+</div>
```
ptrelease, line 175: Because I'm missing it, how are you rendering the text fields so that they appear _next_ to their checkboxes created in the initial loop?

db/migrate/20210406074312_create_heritage_designations.rb
```diff
@@ -0,0 +1,8 @@
+class CreateHeritageDesignations < ActiveRecord::Migration[6.1]
+  def change
+    create_table :heritage_designations, id: :uuid do |t|
+      t.string :designation
+      t.timestamps
+    end
+  end
+end
```
ptrelease, line 8: Having the id as an integer rather than uuid would make support much easier.  I can remember an integer when flicking between tables.  A uuid less so.

ptrelease: Retracting.  Uuids more resilient in the long term - in case somebody deletes some rows - then switches on auto-increment - then the wrong designations will be shown.

db/migrate/20210406074314_add_heritage_designations.rb
```diff
@@ -0,0 +1,61 @@
...
+    execute "delete from heritage_designations where designation = 'grade_1_listed_park_or_garden';"
+    execute "delete from heritage_designations where designation = 'grade_2_star_listed_park_or_garden';"
+    execute "delete from heritage_designations where designation = 'grade_2_listed_park_or_garden';"
+    execute "delete from heritage_designations where designation = 'protected_wreck_site';"
+    execute "delete from heritage_designations where designation = 'national_historic_organ_register';"
+    execute "delete from heritage_designations where designation = 'site_of_special_scientific_interest';"
+    execute "delete from heritage_designations where designation = 'local_nature_reserve';"
+    execute "delete from heritage_designations where designation = 'other';"
+  end
+end
```
ptrelease, line 61: Consider populating this directly after creating the designation table and before the link table.  Because these values go in after the link table, I can't roll the link table back.

app/controllers/funding_application/gp_open_medium/heritage_designations_controller.rb
stuartmccoll: Nope - the database column is `hd_grade_1_description`.

db/migrate/20210406074314_add_heritage_designations.rb
stuartmccoll: Fixed.

app/views/partials/funding_application/_heritage_designations_question.html.erb
stuartmccoll: Fixed.

app/views/partials/funding_application/_heritage_designations_question.html.erb
stuartmccoll: The `textarea` markup is directly beneath it's corresponding `checkbox`.

We loop through all of the heritage designations, and inside the loop we've also got _all_ of the render code for _every_ `textarea` - however, they only trigger conditionally, like this:

```
for all HeritageDesignations do | hd |

checkbox

if hd.designation == 'x' then render textarea

end
```

ptrelease: Got it 👍

ptrelease (approved):

PR #747, opened 2021-04-26
Added filtering to Sentry JavaScript initialisation

stuartmccoll:
## Proposed changes

This commit adds some default filtering to the initialisation of Sentry for reporting on JavaScript errors.

It's safe for us to ignore the 'ResizeObserver loop limit exceeded' error, as this is an error irrelevant to the JavaScript code within our application, but one that we've seen hit Sentry in production.

See https://stackoverflow.com/questions/63020810/what-is-best-way-in-javascript-to-stop-throwing-resizeobserver-loop-limit-excee?noredirect=1&lq=1 for more information.

The rest of the values within the ignoreErrors array, and those within the blacklistUrls array are taken from a list published on the Sentry website under the 'filtering' section.

See https://docs.sentry.io/platforms/javascript/configuration/filtering/#decluttering-sentry for more information.

By enabling this filtering, we should hopefully reduce the amount of Sentry alerts that we've seen in production which have been related to Chrome extension errors, for example.

## Further comments

The blacklistUrls key is renamed to denyUrls in Sentry versions 5.17.0 and above. We're some way behind this at the moment, so will have to rename this key to denyUrls when we bring the package up-to-date.

## Screenshot (if applicable)

N/A.

## Checklist


ptrelease (approved):

ptrelease:
Appreciate the links 👍

PR #755, opened 2021-04-28
 Bug/additional medium welsh fixes

ptrelease:
This addresses some Welsh translation omissions that weren't there for medium go-live.  
A card has been created to address the remaining few issues.


config/locales/address/address_details/en.yml
```diff
@@ -0,0 +1,22 @@
+en-GB:
+  address_details:
+    page_title: "Address details"
+    address_details: "address details"
```
stuartmccoll, line 4:
```suggestion
    address_details: "Address details"
```

PR #805, opened 2021-07-13
Feature/content styling changes to summary pages

ptrelease:
## Proposed changes

Have created a new layout for the summary pages. Exactly like the 'default' application.html.erb layout, but uses the whole page. Drawback is that we can't used the right hand side for 'related link' style content on those pages in the future.


ptrelease:
Also made content changes.

PR #825, opened 2021-09-24
added condition to only show payments option on tasks controller when award < 100k

ptrelease:
## Proposed changes

Applicants querying the payment option, when not currently available for awards > 100k.

Removing the option on the Tasks page, to reduce support calls.

Manually tested.  Also manually tested 3-10k top confirm payment option shows when it should.


ptrelease:
Had intended to include some content changes in this PR - but decided against it.

app/views/funding_application/tasks/show.html.erb
```diff
@@ -106,102 +106,107 @@
 
   </li>
 
-  <li class="nlhf-list__item nlhf-task__item">
-
-    <% if @first_payment_not_started && !@legal_agreement_in_place %>
+  <% # todo: remove when payments possible for medium applications > 100k %>
```
Gubbsy, line 109: Is this still a TODO? It looks like it's been done below?

Gubbsy:
Looks good!

PR #829, opened 2021-09-28
updated content with new agreements bullet point

ptrelease:
## Proposed changes

Added content on the agreement in the word docs and the html pages.


Gubbsy (approved):
All good!

PR #853, opened 2021-10-27
October 2021 content updates

abbiefoxton:
A batch of changes related to tickets #3407 and #3419.

One change includes the addition of a 'your application' bullet point. The code for the bullet point may need reviewing before being added to the live service. 

Other changes are mostly either adding in legal content as requested by legal or fixing typos/ punctuation.


config/locales/funding_application/agreement/check_project_details/en.yml
```diff
@@ -36,8 +36,9 @@ en-GB:
         item_3: "any Additional Grant Conditions, if applicable,"
         item_4_html: "%{href}. If there are any inconsistencies between the Standard Terms of Grant and the Terms and conditions, the %{href} will take precedence,"
         item_4_href: "Standard Terms of Grant"
-        item_5_html: "%{href}."
-        item_5_href: "Receiving a Grant guidance"
+        item_5_html: "%{href}"
+        item_5_href: "Receiving a Grant guidance,"
+        item_6: "Your Application."
       you_must_agree: "You must read and agree to all parts of your Grant Contract before we can pay your Grant."
```
ptrelease, line 42: @abbiefoxton These content changes are fine, but point 6 will not be shown on the page unless there is some corresponding html that refers to it.  Happy to follow up with a chat.

ptrelease: Discussed with Abbie and added HTML

PR #856, opened 2021-11-04
Create PTS form record to upload docs to and abstract shared API help…

Gubbsy:
## Proposed changes

Create PTS form in SF upon submitting an application and attach documents to new record


app/helpers/salesforce_api_helper.rb
```diff
@@ -0,0 +1,88 @@
...
+
+      Rails.logger.debug(
+        "Finished upserting ContentVersion file for #{type} in Salesforce"
+      )
+
+    end
+
+  end
+
+end
```
ptrelease, line 88: newline

app/helpers/permission_to_start_helper.rb
```diff
@@ -400,13 +400,17 @@ def upload_salesforce_pts_files(salesforce_experience_application)
         sf_api.create_file_in_salesforce(
           stat_perm_file,
           "Evidence of statutory permission or licence -#{stat_perm_file.blob.filename}",
-          case_id
+          pts_form_record_id
```
ptrelease, line 403: What is the strategy for retries when uploading files to Salesforce?  

Seems sensible that create_file_in_salesforce handles the retries.

At the moment files uploads are a bit annoying because they're not idempotent - an issue we haven't solved yet.  So ideally we would want each file to retry in the event of a problem, but if it exhausted its retries, then we would want the remaining files to upload, then raise something in Sentry.  Ideally this wouldn't block the applicants submission (otherwise they will click submit again),  and we could follow up in a support ticket.

app/helpers/permission_to_start_helper.rb
```diff
@@ -317,7 +317,7 @@ def update_statutory_permissions_or_licences_json(
   # TODO: Attach to a Large_Grant_Permission_To_Start record, not case
```
ptrelease, line 317: Todo comment left

app/helpers/permission_to_start_helper.rb
ptrelease: Gubbsey has explained how the retries work.

ptrelease (approved):

PR #864, opened 2021-11-09
Fix/identified bugs

Gubbsy:
## Proposed changes

Fix accessibility issues across pages


app/views/salesforce_experience_application/approved_purposes/show.html.erb
```diff
@@ -108,7 +112,6 @@
         )%>
 
       </div>
-
-  <% end %>
-
+    <% end %>
+  </fieldset>
```
ptrelease, line 116: the fieldset and div don't line up here - switch them?

ptrelease:
The git format didn't make it easy to compare the html, so instead I went into each file and checked everything looked OK.  Other than the comment above - it did.

I ran the journey through, only one Wave alert now as you said.  

Consider:

- removing the whitespace - as a lot has been added.
- if you see ```@salesforce_experience_application.errors.any?``` fix the indentation in the few places where it is wrong

Must fix:

- All the delete buttons on file upload forms redirect the journey back to the agreed-costs-documents view.  Have looked into why this is - so hopefully straightforward.

When you've fixed - can you give the journey a thorough test - from the perspective of functionality not accessibility.  There has been a lot of change - so want to check all the errors links and flows still work as they did.

ptrelease (approved):

PR #867, opened 2021-11-11
Fix/print pts form

Gubbsy:
## Proposed changes

Design fixes implemented to pts form prtining


app/assets/stylesheets/helpers/_all.scss
```diff
@@ -11,9 +11,14 @@
     font-style: italic;
   }
 
+  .nlhf-summary, .nlhf-summary__header{
+    color: black !important;
+  }
+
```
ptrelease, line 17: This header is used throughout the other application journeys too.  

Are we prepared to make the same changes to small and medium grant applications?  Or should we create a new CSS class and point the print_form at that?

Gubbsy: This should be done when printing all documents across the applications. It removes an accessibility issue when printing where the nlhf-summary text is grey

ptrelease: Sorry - can see now nested under @media print {

ptrelease (approved):

PR #870, opened 2021-11-12
Bug/delete without js

ptrelease:
## Proposed changes

Changes so that the delete file button on PTS upload pages works when JS is off.


Gubbsy (approved):
All happy bar the notes raised over slack with regards to radio selection.

Gubbsy (approved):
All good!

PR #881, opened 2021-12-06
Fix/legal agreements print fix

Gubbsy:
## Proposed changes

change legal agreements page to download PRint of HTML, as opposed to the existing word document that is presented to the user.


app/views/partials/funding_application/legal_agreements/_terms_signatory.html.erb
```diff
@@ -3,9 +3,35 @@
...
+      </script>
+    <% end %>
+  </p>
+
+  <noscript>
+    <div class="govuk-warning-text no-print">
+      <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+      <strong class="govuk-warning-text__text">
+        <span class="govuk-warning-text__assistive">Warning</span>
+        Download this form using your browser's print functionaility. Javascript is disabled on your device.
```
ptrelease, line 31: Will add the translation.

ptrelease (approved):

PR #885, opened 2021-12-13
Add in file size check

Gubbsy:
## Proposed changes

Fix issue with zero kb uploads throwing error uploading to SF


app/helpers/salesforce_api_helper.rb
```diff
@@ -68,18 +68,24 @@ def insert_salesforce_attachment(
...
+        )
+
+        Rails.logger.debug(
+          "Finished upserting ContentVersion file for #{type} in Salesforce"
+        )
+      else
+        Rails.logger.debug(
+          "Cannot uplaod empty file #{file.blob.filename} in Salesforce - file is of size 0 bytes"
+        )
+      end
```
ptrelease, line 88: I'd definitely make this a Rails.logger.error message - so we see what happened in Prod.

I'd also raise an exception so that the app stops the journey.  For now that will mean an unhelpful rails error - but soon it will be a better 500 page.  If we get more of these we could consider a proper error page - or consider in file validation?

ptrelease: This feels like a better solution that would mean we would show an error message to the applicant: https://stackoverflow.com/questions/6440943/custom-validation-for-file-size-in-rails-before-being-uploaded

Gubbsy: The problem is that we don't open the file and check the size until the upload stage. We will either need to implement this validation technique you linked across all file uploads within FFE (which I am happy to do, but I feel like this is a larger and separate task in itself that will require input from UCD), or we apply the fix I just create which will be applied across all instances.

ptrelease: True.  In which case, disregard my last comment and focus one the first one which has two tasks:

- Change the log level from debug to error - se we can see in the prod logs
- Raise a file exception with the message that filename x has a file size of 0.  Just saves us the trouble of debugging this again when the problem happens next time.

app/jobs/upload_pts_to_salesforce_job.rb
```diff
@@ -54,16 +54,17 @@ def perform(
...
         end
       end
     rescue Exception => e
       raise SalesforceFileUploadError.new("File upload failed for file #{file.blob.filename} on Salesforce Case ID / 
-          #{salesforce_reference} with error #{@e.message}")
+          #{salesforce_reference} with error #{e.message}")
       
-      Rails.logger.debug("File upload failed for file #{file.blob.filename} on Salesforce Case ID / 
-        #{salesforce_reference} with error #{@e.message}")
+      Rails.logger.error("File upload failed for file #{file.blob.filename} on Salesforce Case ID /
```
ptrelease, line 66: Log line needs to go before exception to be written

app/helpers/salesforce_api_helper.rb
```diff
@@ -52,41 +53,50 @@ def insert_salesforce_attachment(
...
 
+        if file.blob.byte_size > 0 
+          client.insert!(
+            'ContentVersion',
+            title: "#{type} File",
+            description: description,
+            pathOnClient: file.blob.filename,
+            versionData: file_upload,
+            FirstPublishLocationId: salesforce_reference
+          )
```
ptrelease, line 81: Minor - but we could make this check before the ```file_upload = Restforce::UploadIO.new(f.path, file.content_type)``` to save the bother of a network call if the file size is zero?

Gubbsy: No worries, I'll change that now

Gubbsy: We need this first for the active job to initialise correctly, tried moving but it bums out

ptrelease (approved):

PR #888, opened 2021-12-17
Add in validation rule for maximum age and move error text to align with gov style

Gubbsy:
Add in validation rule for maximum age and move error text to align with gov style to avoid conflicting with SF during application upload.


app/models/user.rb
```diff
@@ -61,9 +61,13 @@ def date_of_birth_is_date_and_in_past?
...
       errors.add(:date_of_birth, 'Date of birth must be a valid date')
     else
-      unless Date.new(self.dob_year.to_i, self.dob_month.to_i, self.dob_day.to_i).past?
+      dob_provided = Date.new(self.dob_year.to_i, self.dob_month.to_i, self.dob_day.to_i)
+      unless dob_provided.past?  
         errors.add(:date_of_birth, 'Date of birth must be in the past')
       end
+      unless dob_provided > Date.new(1910,1,1)
+        errors.add(:date_of_birth, 'Date of birth cannot be earlier than 1 January 1910')
+      end
```
ptrelease, line 70: @Gubbsy Fine with this change in principle.  However it needs Welsh.  If you want to merge now, and create a follow up card for Welsh, happy to do that.

Also d.o.b of 1st Jan 1910 actually causes the error - so worth tweaking that.

Gubbsy: Good point, yeah. I'll merge and create another follow on the card :)

PR #890, opened 2021-12-22
Add in user authentication before action

Gubbsy:
Add in user authentication before user data retrieval to avoid unhandled exceptions being thrown


app/controllers/pre_application/expression_of_interest/previous_contact_name_controller.rb
```diff
@@ -3,6 +3,7 @@ class PreApplication::ExpressionOfInterest::PreviousContactNameController < Appl
   include ObjectErrorsLogger
   include OrganisationHelper
   include PreApplicationContext
+  before_action :authenticate_user!
```
ptrelease, line 6: PreApplicationContext contains an  authenticate_user!, so this isn't needed.

app/controllers/pre_application/project_enquiry/previous_contact_name_controller.rb
```diff
@@ -3,6 +3,7 @@ class PreApplication::ProjectEnquiry::PreviousContactNameController < Applicatio
   include ObjectErrorsLogger
   include OrganisationHelper
   include PreApplicationContext
+  before_action :authenticate_user!
```
ptrelease, line 6: PreApplicationContext contains an  authenticate_user!, so this isn't needed.

PR #901, opened 2022-02-08
changed sign out button to use translation

ptrelease:
## Proposed changes

Changed signout buttons to use translations where it is missing


Gubbsy (approved):
LG!

PR #910, opened 2022-02-15
Feature/ammend welsh links

Gubbsy:
## Proposed changes

Add in welsh counterparts for all external service links.


app/views/funding_application/gp_project/start/show.html.erb
```diff
@@ -85,7 +85,7 @@
           "start.read_about_what_we_fund_html",
           href: link_to(
                   t("start.read_about_what_we_fund_text"),
-                  "https://www.heritagefund.org.uk/funding/check-what-we-fund",
+                  "#{I18n.locale == :cy ? 'https://www.heritagefund.org.uk/cy/node/110862' : 'https://www.heritagefund.org.uk/funding/outcomes' }",
```
ptrelease, line 88: Can you double-check this one.  It's a bit confusing as the /check-what-we-find link in the original code resolves to https://www.heritagefund.org.uk/funding.  Which describes what we fund.  And there is a Welsh equivalent.

However, you have changed top the outcomes page here.

Gubbsy: You're right, I'll fix that now

app/views/partials/funding_application/legal_agreements/_terms.html.erb
```diff
@@ -196,7 +196,7 @@
           'agreement.organisations_must_list.item_2_html',
           href: link_to(
             t('agreement.organisations_must_list.item_2_href'),
-            'https://www.gov.uk/government/publications/the-7-principles-of-public-life/the-7-principles-of-public-life--2',
+            "#{I18n.locale == :cy ? 'https://gov.wales/sites/default/files/publications/2020-12/ATISN%2014529%20Doc%2003d.pdf' : 'https://www.gov.uk/government/publications/the-7-principles-of-public-life/the-7-principles-of-public-life--2'}",
             class: "printable-link"
             )
```
ptrelease, line 201: Did this come from Liam?  Its hosted on a Gov website.  But the content doesn't exactly match the GOV.UK file.

app/views/funding_application/gp_project/start/show.html.erb
Gubbsy: Fixed

ptrelease: in /support/show.html you link to 'what-we-fund' which looks better.

app/views/partials/funding_application/legal_agreements/_terms.html.erb
Gubbsy: Yes, Liam gave me the welsh counterpart to this link on the welsh gov website, it's in the chain of emails we had.

ptrelease (approved):

ptrelease (approved):

PR #912, opened 2022-02-16
Fix/welsh links and related

Gubbsy:
## Proposed changes

Fix outstanding links for welsh


app/views/funding_application/gp_project/volunteers/show.html.erb
```diff
@@ -83,10 +83,11 @@
...
-  <%= render(ButtonComponent.new(element: 'input')) %>
-</form>
+<%=
+  render(
+    ButtonComponent.new(
+      element: 'a',
+      href: funding_application_gp_project_evidence_of_support_path
+    )
+  )
+%>
```
ptrelease, line 93: If we're ditching the hint, this is fine.  Just remove the text from the yml to tidy that up.

If you go back into this file - add the newline at the end.

Gubbsy: I've added the hint text so it lines up with medium. I changed it from a form action before as it removed the translation params from the URL on the following pages.

app/views/partials/funding_application/_capital_works_question.html.erb
```diff
@@ -62,27 +62,80 @@
...
+              %>
+            <% elsif model_object_name == "open_medium"%>
+              <%= 
+                t(
+                  'capital_works.examples_physical_work_medium_html',
+                  examples_gp_om_1_guidance: link_to(
+                    t("capital_works.examples_gp_#{ model_object_name }_1_guidance"),
+                    "#{I18n.locale == :cy ? 
+                      'https://www.heritagefund.org.uk/cy/funding/receiving-grant-guidance-ps10000-ps100000#heading-10' : 
+                        'https://www.heritagefund.org.uk/funding/national-lottery-grants-heritage-10k-250k/receiving-grant-100k#heading-10'}",
```
ptrelease, line 99: I think the link on line 98 is wrong?  In expect the values for the links to match for lines 98 and 99.

app/views/partials/funding_application/_capital_works_question.html.erb
```diff
@@ -62,27 +62,80 @@
...
+                    {
+                      class: 'govuk-link',
+                      target: '_blank'
+                    }
+                  ),
+                  examples_receiving_a_grant_guidance_medium_1: link_to(
+                    t('capital_works.examples_receiving_a_grant_guidance_medium_1'),
+                    "#{I18n.locale == :cy ? 
+                      'https://www.heritagefund.org.uk/cy/funding/receiving-grant-guidance-ps10000-ps100000' : 
+                        'https://www.heritagefund.org.uk/funding/national-lottery-grants-heritage-10k-250k/receiving-grant-100k'}",
```
ptrelease, line 119: Again, check the links

app/views/partials/funding_application/_capital_works_question.html.erb
```diff
@@ -62,27 +62,80 @@
...
+                        'https://www.heritagefund.org.uk/funding/national-lottery-grants-heritage-10k-250k/receiving-grant-250k'}",
+                    {
+                      class: 'govuk-link',
+                      target: '_blank'
+                    }
+                  )
+                )
+              %>
+            <% end %>
+
```
ptrelease, line 138: Generally speaking, this is a complex way of doing it.  Introduces a lot of logic on the view that we will need to write Jest tests for when there is time.

Would prefer that the logic lives server-side on the show method.  Within that method you could have:

```
@receiving_grant_guidance_link = some ternary logic that looks at funding_app type
```
Then you can refer to that instance variables in the forms and partials.  And test with RSPEC later.

ptrelease: https://github.com/heritagefund/funding-frontend/blob/947b7d03f378b4656a77225790f6aac91f4d37cd/app/helpers/funding_application_helper.rb#L95

app/views/partials/funding_application/_capital_works_question.html.erb
ptrelease: Adam explained this is right

app/views/partials/funding_application/_capital_works_question.html.erb
ptrelease: Adam explained this is right

app/views/partials/funding_application/_capital_works_question.html.erb
ptrelease: Content needs further review.  So we will consider implementing that way when the content changes.

PR #914, opened 2022-02-17
Feature/late legal agreements welsh

ptrelease:
## Proposed changes

Late Welsh content changes for legal agreements


config/locales/en.yml
```diff
@@ -17,6 +17,8 @@ en-GB:
       submit: "Submit"
       i_agree: "I agree"
       sign_out: "Sign out"
+      sign_in: "Sign in"
```
Gubbsy, line 20: Are we missing the welsh counterparts for these?

config/locales/cy.yml
```diff
@@ -17,6 +17,8 @@ cy:
       submit: "Cyflwyno"
       i_agree: "Rwy'n cytuno"
       sign_out: "Allgofnodi"
+      sign_in: "Mewngofnodi"
+      your_account: "Eich cyfrif"
   generic:
```
ptrelease, line 22: Here are the Welsh counterparts

Gubbsy (approved):

PR #935, opened 2022-03-18
Feature/additional grant conditions

ptrelease:
## Proposed changes

Added basic additional grant conditions page as a continuation of the arrears payment journey.  Onward journey to follow.  Error message validation doesn't follow design at the moment.  This will need rework.


Gubbsy (approved):
All good!

PR #952, opened 2022-04-05
Feature/cash contributions select

ptrelease:
## Proposed changes

Added pages to add cash contribution updates as part of the arrears journey.

## Further comments

The ProgressUpdateCashContribution class is responsible for validating and storing updates provided
about Cash Contributions so that they can be sent to Salesforce
as part of an arrears payment request.

An object is only created from this class when an applicant
selects a cash contribution to update.  This decision is
stored using the "selected_for_update" key nested in the
["cash_contribution"]["records"] part of
the answers_json field of the progress_updates table.
As such, the ["cash_contribution"]["records"] key should be
checked to see what ProgressUpdateCashContribution objects
were "selected_for_update" before sending information to Salesforce, as
an applicant can change their minds on what cash_contribution they update
but no ProgressUpdateCashContribution objects (or pg rows)
are destroyed once created.


app/helpers/progress_and_spend_helper.rb
```diff
@@ -126,7 +127,141 @@ def toggle_risk_validation(risk, params)
...
+          salesforce_income.Amount_you_have_received__c
+
+        cash_contribution.display_text =
+          salesforce_income.Description_for_cash_contributions__c
+
+        cash_contribution.save!
+
+        # update this cash_contribution next, unless one chosen
+        next_cash_contribution_id = cash_contribution.id unless \
+          next_cash_contribution_id.present?
```
Gubbsy, line 195: What is the purpose of the`unless`? It's not set anywhere other than its initialisation to nil at the start of the function?

Gubbsy (approved):
Looks good

app/helpers/progress_and_spend_helper.rb
ptrelease: The unless isn't strictly necessary.  If you take it away, then it's always the last unfinished record in the loop that is assigned as next_cash_contribution_id.  Practically speaking this tends to mean that the cash contributions journey would process each cash contribution in the opposite order to that shown on the select page.  So I added the unless for usability (IMO).

Gubbsy: Ah I see, all looks good to me then!

PR #953, opened 2022-04-07
coded and tested

ptrelease:
## Proposed changes

Fixed payment bug, where an organisation using the same account details a second time on a new application encountered an error.


Gubbsy (approved):
Looks Good

PR #954, opened 2022-04-07
fixes three issues with additional grant conditions

ptrelease:
## Proposed changes

Fixed dangerous ordering issue that could mean an applicant that was using was clicking back could see the answers were mixed up, or worse still resubmit with the answers mixed.  Fixed by specifying the order that the associated ProgressUpdateAdditionalGrantCondition instances are loaded in from the ProgressUpdate model.

Opted to remove ProgressUpdateAdditionalGrantCondition instances with no answers.  Needed to submit updates.

Included an "additional_grant_condition" key for answers_json.  Consistent with rest of journey and useful for support.


Gubbsy (approved):
Approved

PR #956, opened 2022-04-08
Feature/check your answers

Gubbsy:
## Proposed changes

Add in check your answers page which is dynamically generated from answers given in the form.
Links to change answers are provided at end of each answers section.

Also added in status enum that is set at the start and end of the journey and used to take the user to either the start of the journey if not_started or check your answers if completed.


ptrelease:
Looks great.  Only comment is that the controllers have many lines over 80 chars.  Would be great to break those down possible to avoid later refactoring when we add linting to the pipeline.

ptrelease (approved):

PR #960, opened 2022-04-12
Feature/prjct updt clear unused data

Gubbsy:
## Proposed changes

Add method call to upload progress update to SF (not implemented) in helper and clear out unused data based on answers json.


app/helpers/progress_and_spend_helper.rb
```diff
@@ -346,5 +346,89 @@ def salesforce_approved_purposes(funding_application)
...
+  # Method responsible for orchestrating upload
+  # of prpogress update data to Salesforce
+  #
+  # @param [ProgressUpdate] progress_update An instance of
+  #                                                 ProgressUpdate
+  def upload_progress_update(progress_update)
+
+    # TODO: Upload details to salesforce with client
+
+    progress_update.answers_json.each do | field |
```
ptrelease, line 359: A nicer way to do this is by referring to |key, value|

Then you refer to key and value instead of field[0] and field[1] respectively

ptrelease: or field/flags in this case

app/helpers/progress_and_spend_helper.rb
```diff
@@ -346,5 +346,89 @@ def salesforce_approved_purposes(funding_application)
...
+      end
+    when 'procurements'
+      if flags['has_procured_goods'] == "true"
+        flags['has_procurement_report_evidence'] == 'true' ?
+          progress_update.progress_update_procurement.destroy_all : 
+          progress_update.progress_update_procurement_evidence.destroy_all
+      else
+        progress_update.progress_update_procurement.destroy_all  
+        progress_update.progress_update_procurement_evidence.destroy_all
+      end
```
ptrelease, line 403: Based on the code we have now - this is fine - because we validate that these flags are true or false.

Reading this again, makes me a little uneasy that we delete anything with an 'else'. 

This is because we could introduce a bug, or omit validation.  In that scenario, 'else' will fire and we will delete.

Would it be better defensive programming to only delete if flag == 'false'?  

It's slightly more code, but it will mean we keep information when conditions aren't met - rather than deleting the wrong thing.

app/helpers/progress_and_spend_helper.rb
```diff
@@ -346,5 +346,89 @@ def salesforce_approved_purposes(funding_application)
         (funding_application.salesforce_case_id)
 
   end
+
+  # Method responsible for orchestrating upload
+  # of prpogress update data to Salesforce
```
ptrelease, line 351: prpogress

app/helpers/progress_and_spend_helper.rb
```diff
@@ -346,5 +346,89 @@ def salesforce_approved_purposes(funding_application)
...
+    end
+
+  end
   
+  private
+
+  # Method responsible for deleting unused 
+  # prpogress update data.
+  #
+  # @param [field] field String vlaue denoting field in answers JSON
```
ptrelease, line 374: vlaue

Gubbsy: Fixed

app/helpers/progress_and_spend_helper.rb
Gubbsy: Fixed

app/helpers/progress_and_spend_helper.rb
Gubbsy: Actioned

app/helpers/progress_and_spend_helper.rb
Gubbsy: Implemented if check

ptrelease (approved):

PR #964, opened 2022-04-21
Feature/upsert project update

Gubbsy:
## Proposed changes

Upsert all attached documents related to a progress update and as well as creating related sub salesforce data objects that correspond to the information captured throughout the progress update journey. 

Upload of evidence files re-uses API for PTS payments which could be refactored and abstracted at a later date to better reflect its generic application.


app/helpers/progress_and_spend_helper.rb
```diff
@@ -1,5 +1,6 @@
 module ProgressAndSpendHelper
   include SalesforceApi
+  include PtsSalesforceApi
```
ptrelease, line 3: Can remove this inclusion now?

Gubbsy: No, as we are re-using the PtsSalesfoceAPI to send the files. I made a note that we may want to abstract this later down the line, but it would require some refactoring to PTS payments.

app/helpers/progress_and_spend_helper.rb
```diff
@@ -441,19 +442,29 @@ def build_outcomes_hash_from_params(form_params)
...
     progress_update.answers_json.each do | field, flags |
       clear_unused_progress_update_data_items(
         field, 
         flags, 
         progress_update
       )
-    end
+    end 
+
+    upload_evidence_files(progress_update, salesforce_project_update_id)
```
ptrelease, line 467: May be worth moving above the clear_unused_progress_update_data_items function. 

Just means that if clear_unused_progress_update_data_items fails, then the files are already in Salesforce.

app/helpers/progress_and_spend_helper.rb
```diff
@@ -530,4 +541,95 @@ def clear_unused_progress_update_data_items(field, flags, progress_update)
...
+  # Method responsible for upserting any evidence files attached to 
+  # project update (calls file uploader used in PtsSalesforceAPI
+  # which could later be abstarcted to more generic helper)
+  #
+  # @param [ProgressUpdate] progress_update An instance of
+  #                                                 ProgressUpdate
+  # @param [salesforce_project_update_id] Form Id of progress update
+  #                                                  String
+  def upload_evidence_files(progress_update, salesforce_project_update_id)
+    pts_salesforce_api_client= PtsSalesforceApiClient.new
```
ptrelease, line 554: Ah - I see why you included PtsSalesforceApiClient now.

When we were discussing this yesterday, we considered leaving the PtsSalesforceApiClient code where it was (because the PTS stuff will be rebuilt) and reuse the code somewhere common.  To be honest Salesforce_api is the most common - if you would rather copy there than the specialised progress and spend salesforce lib.

I still think that's the best plan, rather than link arrears to PTS.  Doing that means we will have to refactor and retest arrears when we retire PTS.

lib/apis/salesforce/progress_update/progress_update_salesforce_api.rb
```diff
@@ -92,8 +92,157 @@ def get_project_outcome_targets(salesforce_case_id)
 
     end
 
-    private
+  # Method responsible for upserting any project update data models
```
ptrelease, line 95: It's progress update on this comment line.  

It does get a bit confusing referring to progress in FFE and project in SF.  But your var names look good.

lib/apis/salesforce/progress_update/progress_update_salesforce_api.rb
```diff
@@ -92,8 +92,157 @@ def get_project_outcome_targets(salesforce_case_id)
...
+
+        return salesforce_project_update_id
 
+      rescue Restforce::MatchesMultipleError, Restforce::UnauthorizedError,
+        Restforce::EntityTooLargeError, Restforce::ResponseError => e
+
+        Rails.logger.error(
+          "Error upserting project update with ID: #{progress_update.id}"
+        )
+
```
ptrelease, line 237: It would be useful to info log as you successfully complete each upsert.

Something like 'New expiry date successfully uploaded to Salesforce for payment id: 123'.

lib/apis/salesforce/progress_update/progress_update_salesforce_api.rb
```diff
@@ -92,8 +92,157 @@ def get_project_outcome_targets(salesforce_case_id)
...
+        # Raise and allow global exception handler to catch
+        raise
+
+      end
+
```
ptrelease, line 242: Could we log a success message?

app/helpers/progress_and_spend_helper.rb
Gubbsy: I do the clear before so we get rid of any unused evidence files before we do the upload. If this wasn't performed before then we will upload evidence files that are hanging around after the user has changed their answer.

ptrelease:
Uploads: testing happy path:
fa: 3e6ab29f-2788-4534-9185-d59a5d4c046e
photos - tabby
events - pale tabby 
staff - ragdoll
procurement evidence - ginger kitten
new completion date - birthday
perms, licences - tom by wall
risk register - tabby eating
cc1 - no with amount and date.  
cc2 - no with amount and reason. 
vol1 with 1 hr added
scc1 with £1 added

No bugs

Uploads: testing that unused stuff not submitted:
photos - grumpy tabby
events - ginger tom
new staff - long haired tabby
procurement report - tom by wall
procurement detail: supp name, item desc, 01-01-1990, £15000, not lowest tender
GED - Yes
Stat, permissions - ginger kitten
risk register - ragdoll
risk detail: flood, Likelihood - low, Impact low, yes a risk - sandbags.
cc1 only (yes - getting what expected)
no volunteer updates
no secured ncc updates

no bugs

ptrelease:
Tested that no unused files stored FFE side when not required.

ptrelease (approved):

PR #969, opened 2022-04-28
Feature/lower spend add

ptrelease:
## Proposed changes

Added spend add page to the arrears journey.

## Further comments

The form works on the basis that the first item in the spends_to_do array represents the form we are currently updating.  This applies to the show and update.  
The only scenario where this could cause a problem is when an applicant is running the same journey on two tabs or machines.  But even then they could fix errors when reviewing their work on the summary page.
To mitigate this we could pass a unique id to the form as a hidden field, then check this coming back.  Or create the low spend object up front and delete later if required.  Though, both options add significant complexity to the code and do not warrant the effort.

The error messages upon validation don't work well if a string is entered as opposed to a decimal.  Fed back to design team for later iteration.

Noticed the indentation on low_spend_select_controller.rb and what_spend_controller.rb was wrong - so corrected. In these controllers, no other changes were made except changing get_low_spend_amount_from_json to get_spend_threshold_from_json in  low_spend_select_controller.rb.


app/controllers/funding_application/progress_and_spend/payments/what_spend_controller.rb
```diff
@@ -2,78 +2,78 @@ class FundingApplication::ProgressAndSpend::Payments::WhatSpendController < Appl
...
+    payment_request.answers_json['arrears_journey']\
+      ['spend_journeys_to_do'] = selected_array
+
+    payment_request.save
 
-    end
-  
   end
+
+end
```
Gubbsy, line 79: New line

PR #971, opened 2022-04-29
Feature/arrears bank details

Gubbsy:
## Proposed changes

Modify previously unimplemented bank accounts journey to fit in with arrears journey. This included some content adjustment and the addition of evidence pages to align with the design. The arrears journey routes to this bank journey and sets the appropriate data models, linking back to the tasks page upon completion.


app/controllers/funding_application/bank_details/confirm_controller.rb
```diff
@@ -1,72 +1,13 @@
...
-
-  end
-
-  # Method responsible for validating existence of payment details evidence file
-  # before redirecting to the next page in the journey
-  def save_and_continue
-
-    logger.info 'Navigating past payment details confirmation screen '
+    logger.info 'User has confirmed payment details on confirmation screen '
                 "for payment_details ID: #{@funding_application.payment_details.id}"
```
ptrelease, line 9: Go through arrears tracker.  Otherwise this could be wrong when there is more than one payment against the funding application.

config/locales/funding_application/payment_legal_sig_account_evidence/en.yml
```diff
@@ -0,0 +1,18 @@
...
+    evidence_we_accept: "Evidence can be a recent bank statement, cheque, or paying-in slip."
+    bank_statements: "Bank statements need to be from within the last 3 months. Online bank statements need to show the bank logo, website address, account name, number and sort code."
+    evidence_reviewed: "All bank evidence needs to be reviewed and signed by your legal signatories."
+    legal_signatory_evidence_file_label: "Upload evidence of bank account signed by legal signatories"
+    no_evidence_uploaded: "You have not added evidence of your bank account signed by legal signatories"
+    legal_signatory_evidence_file: "Evidence of bank account file, signed by legal signatories"
+    buttons:
+      add_evidence:
+        text: "Add evidence of bank account"
+
```
ptrelease, line 18: tidy up file endings

ptrelease:
Content implies that bank statements (plural) can be uploaded as evidence.

However, only one can be uploaded.  Design doesn't specify - but probably worth checking as this could be a showstopper.

app/models/payment_details.rb
```diff
@@ -83,6 +89,13 @@ def encrypt_account_name?
         :evidence_file,
         I18n.t("activerecord.errors.models.payment_details.attributes.evidence_file.inclusion")
     ) if validate_evidence_file?
+
+    validate_file_attached(
+      :legal_signatory_evidence_file,
+      I18n.t("activerecord.errors.models.payment_details.attributes.legal_signatory_evidence_file.inclusion")
+  ) if validate_legal_signatory_evidence_file?
```
ptrelease, line 96: indentation a little off?

ptrelease:
Tried reusing the new form and noticed two problems.

1) Hit save and continue with no file attached.  Then refresh page - check there is no error.  I have fix if there is.
2) Check summary errors renders in the right place.

app/models/payment_details.rb
Gubbsy: Removed

Gubbsy:
> Tried reusing the new form and noticed two problems.
> 
> 1. Hit save and continue with no file attached.  Then refresh page - check there is no error.  I have fix if there is.
> 2. Check summary errors renders in the right place.

If you have a fix for that error with a refresh that would be great. Regarding the summary errors, Abbie would like the summary to stay where it is when we did our content review. 

I have removed the signatories page mistakenly implemented from the old design.

ptrelease (approved):

PR #972, opened 2022-04-29
Feature/low spend summary

ptrelease:
## Proposed changes

This concludes the low spend journey.  Apart from submitting to Salesforce.

Queried with designers whether we should validate that spends are under the threshold.  Also assumed that we are limiting the table of spend to a single file upload - but have asked interaction designer.

## Proposed changes

Moved the add spend view content into a partial - so it can be used by add and edit.  Renamed the en.yml accordingly.


Gubbsy (approved):
Looks good!

PR #981, opened 2022-05-17
Feature/funding acknowledgement

ptrelease:
## Proposed changes

Added funding acknowledgement page.

## Further comments

Stores acknowledgements as JSON, example below.

Storing as JSON grants flexibility, as these acknowledgements could be
changed, added or removed.

Only submit to Salesforce if 'no_update' is NOT selected.  Then only
submit acknowledgements that were selected.

If 'no_update' is selected, the JSON could be cleaned up, but there
are currently no files associated - so significant storage savings in
doing this.


Gubbsy (approved):
Looks Good!

PR #984, opened 2022-05-19
Feature/submitting payment requests

Gubbsy:
## Proposed changes

Upload payment request spend details to Salesforce. For this, you need to generate the appropriate form__c based on what you are submitting (performed using the `initialise form` method). This also required some changes to the existing progress_update uploaders, referring to the complete_arrears_journey id for the external id of the form__c. 

NOTE: I am aware that the spend API calls have been placed within the progress_update_salesforce_api, upon reflection the spend API calls could/should have been placed within a separate API module. If this is an issue we can either rename the API module to arrears_salesforce_api or abstract the spend calls to a separate API module. 

## Fixes/minor additions
- Add missing additional grant conditions uploader
- Clean up arrears_journey_tacker upon submission
- Redirect users to dashboard if they have submitted (no longer has an arrears journey tracker against funding app)


app/controllers/funding_application/progress_and_spend/submit_controller.rb
```diff
@@ -15,36 +14,36 @@ def update()
...
+            .payment_request.answers_json["bank_details_journey"]["has_bank_details_update"] == "true"
+        end
 
-    end
+        # TODO: Pull down payment amount
+        @payment_amount = 1000
 
-    def submit_to_salesforce()
+        # TODO: Pull down payment percentage
+        @payment_percentage = 85
```
ptrelease, line 38: Just wondering what is stopping us doing this now?

app/controllers/funding_application/progress_and_spend/submit_controller.rb
```diff
@@ -55,9 +54,14 @@ def submit_to_salesforce()
         submitted_on: DateTime.now()
       )
 
+      upload_arrears_to_salesforce(@funding_application, completed_arrears_journey)
+
       @funding_application.completed_arrears_journeys << 
         completed_arrears_journey
```
ptrelease, line 60: So this line seems redundant, as you've already added the references manually, and don't refer to the completed_arrears_journeys collection again in the scope of this function.

If you wanted to be more Railsy, could you change the create to be more like:
```
@funding_application.completed_arrears_journeys.create(
   payment_request_id: arrears_journey_tracker&.payment_request&.id?,
   progress_update_id: arrears_journey_tracker&.progress_update&.id
   submitted_on: DateTime.now()
)

File: app/controllers/funding_application/progress_and_spend/submit_controller.rb [hunk 3]
```diff
@@ -55,9 +54,14 @@ def submit_to_salesforce()
...
 
+      # Remove arrears_journey_tracker
+      @funding_application.arrears_journey_tracker.delete
+
       @funding_application.save
```
ptrelease, line 65: What does a save do here?   Seems as though the delete on line 63 is all you need?

lib/apis/salesforce/progress_update/progress_update_salesforce_api.rb
```diff
@@ -282,6 +348,102 @@ def upsert_project_update(funding_application)
...
+            'External_Id__c',
+            External_Id__c: high_spend.id,
+            Forms__c: salesforce_payment_request_id,
+            Cost_Heading__c: high_spend.cost_heading,
+            Amount__c: high_spend.amount , 
+            VAT__c: high_spend.vat_amount,
+            Date_of_spend__c: 
+              high_spend.date_of_spend&.strftime("%Y-%m-%d"),
+            Description__c: high_spend.description,
+            Spend_level__c: "Spend over £500"
```
ptrelease, line 384: Use: 
```
Spend_level__c: high.spend.spend_threshold
```

I store this as its different when we go to large stuff

lib/apis/salesforce/progress_update/progress_update_salesforce_api.rb
```diff
@@ -282,6 +348,102 @@ def upsert_project_update(funding_application)
...
+        else
+          # Raise and allow global exception handler to catch
+          raise
+        end
+
+      end
+
+    end
+
     # Method responsible for upserting any approved purpose models
```
ptrelease, line 447: Use payment_request_salesforce_api.rb fo rpayment stuff.

lib/apis/salesforce/progress_update/progress_update_salesforce_api.rb
Gubbsy: Done

app/controllers/funding_application/progress_and_spend/submit_controller.rb
Gubbsy: Resolved

app/controllers/funding_application/progress_and_spend/submit_controller.rb
Gubbsy: Resolved

lib/apis/salesforce/progress_update/progress_update_salesforce_api.rb
```diff
@@ -117,32 +117,84 @@ def upsert_document_to_salesforce(
...
     # to their counter parts in SF
     #
     # @param [FundingApplication] funding_application An instance of
     #                                                 FundingApplication
-    #
-    def upsert_project_update(funding_application)
+    # @param [String] string id for SF Form to upsert against
+    # 
+    # @param [completed_arrears_journey] completed_arrears_journey 
+    #                                                 An instance of
```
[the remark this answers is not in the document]
Gubbsy: Done

lib/apis/salesforce/progress_update/progress_update_salesforce_api.rb
```diff
@@ -282,6 +348,102 @@ def upsert_project_update(funding_application)
 
     end
 
+    # Method responsible for upserting any progress update data models
+    # to their counter parts in SF
+    #
+    # @param [FundingApplication] funding_application An instance of
+    #                                                 FundingApplication
+    # @param [String] string id for SF Form to upsert against
```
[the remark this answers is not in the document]
Gubbsy: Done

lib/apis/salesforce/progress_update/progress_update_salesforce_api.rb
Gubbsy: Actioned

app/controllers/funding_application/progress_and_spend/submit_controller.rb
```diff
@@ -55,9 +54,14 @@ def submit_to_salesforce()
         submitted_on: DateTime.now()
       )
```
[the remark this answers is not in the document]
Gubbsy: Agreed, have added migration and set values upon completion of upload

ptrelease (approved):
Tested file uploader

app/controllers/funding_application/progress_and_spend/tasks_controller.rb
```diff
@@ -93,5 +97,30 @@ def retrieve_project_info
   
     end
 
+    def submit_to_salesforce()
+      completed_arrears_journey = @funding_application.completed_arrears_journeys.create(
+        payment_request_id: arrears_journey_tracker&.payment_request&.id,
+        progress_update_id: arrears_journey_tracker&.progress_update&.id,
+        submitted_on: DateTime.now()
+     )
```
ptrelease, line 102: We occasionally have problems when the Salesforce response doesn't come back properly.  

In those scenarios - the user tries again later.

To stop multiple completed_arrears_journey instances from being created, can you check first, and create if not found.

app/controllers/funding_application/progress_and_spend/tasks_controller.rb
```diff
@@ -93,5 +97,30 @@ def retrieve_project_info
...
+    end
+
+    def arrears_journey_tracker
+      @funding_application.arrears_journey_tracker
+    end
+
+    def has_completed?
+        CompletedArrearsJourney.where(payment_request_id:  @funding_application.arrears_journey_tracker&.payment_request_id).present? ||
+        CompletedArrearsJourney.where(progress_update_id:  @funding_application.arrears_journey_tracker&.progress_update_id).present?
+    end
```
ptrelease, line 126: Go through funding_application.completed_arrears_journey if you can.  I would expect it to be more performant as the underlying SQL should include more criteria.  I can't think of a scenario, even in dev, where we could end up associating the wrong pay/prog updates to the wrong application - but best practice to be safe.

ptrelease: Also, fix the bug where this check returns a false positive when matching on a row with null as the id.

app/controllers/funding_application/progress_and_spend/submit_controller.rb
Gubbsy: Will update in later PR

app/controllers/funding_application/progress_and_spend/tasks_controller.rb
```diff
@@ -53,7 +57,7 @@ def show()
     end
   
     def update()
-  
+      submit_to_salesforce if params.has_key?(:submit_button)
```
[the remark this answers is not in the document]
Gubbsy: This is to ensure that if the user was to attempt a URL hack and try and post from a browser without the button press it bums out. I just wanted to make sure they definitely click the button, which is only available upon completion of all the tasks.

app/controllers/funding_application/progress_and_spend/tasks_controller.rb
Gubbsy: Removed this check and replaced with an existence check of areas journey tracker (indicating they can submit and that they haven't previously submitted successfully)

app/controllers/funding_application/progress_and_spend/tasks_controller.rb
Gubbsy: We now check to see if a completed journey object exists with these IDs and return if it does (this would be a re-submission after an unsuccessful submission scenario), else create a new object.

ptrelease (approved):

PR #985, opened 2022-05-20
check addde

ptrelease:
## Proposed changes

Quick check to see if Salesforce already has bank details stored.  If it does not, we don't ask the user if the bank details have changed.


Gubbsy (approved):
Looks good

PR #986, opened 2022-05-20
Reinstated old SF query - created optimised version for arrears only

ptrelease:
## Proposed changes

Reverted generic project details soql query.  Split out additional fields into a new query.

This is for performance, and to let us deploy to production without an accompanying SF changeset.


Gubbsy (approved):
Happy

PR #988, opened 2022-05-26
Feature/submit bank details

Gubbsy:
## Proposed changes

Upload bank details to SF re-using existing salesforce API methods, based on whether a given set of bank details already live in SF and are voided or not verified.


app/helpers/progress_and_spend_helper.rb
```diff
@@ -571,29 +583,117 @@ def upload_progress_update(client, funding_application,
...
+      account_num_sort_code = 
+        payment_details.decrypt_account_number +
+          payment_details.decrypt_sort_code
+
+      exisitng_account_details =
+       payment_request_client
+        .exisitng_bank_details_status_in_salesforce(
+          salesforce_account_id, 
+          account_num_sort_code
+        )
```
ptrelease, line 639: typo in 'existing'

app/helpers/progress_and_spend_helper.rb
```diff
@@ -571,29 +583,117 @@ def upload_progress_update(client, funding_application,
...
+
+      if exisitng_account_details[:details_aleardy_exist] && 
+          !exisitng_account_details[:void] &&
+            exisitng_account_details[:verified]
+
```
ptrelease, line 644: typo in 'already'

app/helpers/progress_and_spend_helper.rb
```diff
@@ -571,29 +583,117 @@ def upload_progress_update(client, funding_application,
...
+        )
+
+      if exisitng_account_details[:details_aleardy_exist] && 
+          !exisitng_account_details[:void] &&
+            exisitng_account_details[:verified]
+        
+        # Do not create duplicate bank account details as matching, 
+        # verified and valid details already exsit
+        salesforce_bank_account_id = exisitng_account_details[:account_id]
+      else
```
ptrelease, line 648: typo in exisitng_account_details

app/helpers/progress_and_spend_helper.rb
ptrelease: This logic looks fine, but seems unnecessary.  I think the SOQL query can do this for us.  Will try and create a demo query later in this PR.

lib/apis/salesforce/payment_request/payment_request_salesforce_api.rb
```diff
@@ -175,7 +172,56 @@ def org_has_bank_account_in_salesforce(salesforce_account_id)
...
+              
+        response_hash[:details_aleardy_exist] = true
+        response_hash[:account_id] = restforce_response.first[:Id]
+        response_hash[:void] = restforce_response.first[:Void__c]
+        response_hash[:verified] = restforce_response.first[:Verified__c]
+      end
+
+      response_hash
+  
+    end
```
ptrelease, line 224: We only want to check the _latest_ bank account with matching criteria.  So the second query isn't right.  As the result set is unsorted by date - any of the bank account details - regardless of date could be the first row.

With hindsight I think we could correct the SOQL and merge the queries to make them more efficient:

```
SELECT id FROM Bank_Account__c where Organisation__c = '0013M000012pNa2QAE' and Account_Number_Sort_Code__c = '19191919191919' and Void__c = false and Verified__c = true order by CreatedDate desc Limit 1
```
This also means you need no rails logic.  If you get a row from this SOQL, then you can use it.  Otherwise create a new record.

All the logic we need lives in the query - so we have a more efficient SF query, returning less data, and requiring no rails unit testing.

Only disadvantage I see is that you're forced to create the data in SF to manually test (no bad thing).

lib/apis/salesforce/salesforce_api.rb
```diff
@@ -1730,44 +1726,50 @@ def upsert_bank_account_details(funding_application)
...
+          Account_Number_Sort_Code__c: funding_application.payment_details.decrypt_account_number \
+            + funding_application.payment_details.decrypt_sort_code
         )
 
         Rails.logger.info(
-          'Created a form bank account junction record in Salesforce with ' \
-          "reference: #{salesforce_form_bank_account_id}"
+          'Created a bank account record in Salesforce with ' \
+          "reference: #{salesforce_bank_account_id}"
         )
```
ptrelease, line 1851: The message for your upsert and insert functions are the same.  Could we replace 'create' with 'upsert' or 'insert' as appropriate - so the logs are better?

app/helpers/progress_and_spend_helper.rb
```diff
@@ -638,18 +638,21 @@ def upload_bank_details(form_id, salesforce_api_client,
           account_num_sort_code
         )
 
-      if exisitng_account_details[:details_aleardy_exist] && 
-          !exisitng_account_details[:void] &&
-            exisitng_account_details[:verified]
-        
+      if exisitng_account_details.present?
```
ptrelease: You could correct spelling of existing if you wanted.

lib/apis/salesforce/payment_request/payment_request_salesforce_api.rb
```diff
@@ -179,50 +179,77 @@ def org_has_bank_account_in_salesforce(salesforce_account_id)
     # @param [String] unique account num/sort code combination that identifies bank account
     #                              details associated with organisation.   
     # 
-    #  @return [Hash] hash of data associated with bank account including: if the account exists already,
-    #                             it's void and verified status, alongside any existing account id. 
+    # @return [<Restforce::SObject>] A Restforce  response object containing
+    #                                               with query results
     # 
     def exisitng_bank_details_status_in_salesforce(
       salesforce_account_id, account_num_sort_code)
```
ptrelease: You could correct spelling of existing if you wanted.

app/helpers/progress_and_spend_helper.rb
ptrelease: Already changed in subsequent commit

lib/apis/salesforce/salesforce_api.rb
ptrelease: Already changed in subsequent commit

ptrelease:
Logic looks sound.  There is a problem with the retries being managed by:

- upload_bank_details (progress and spend)
- upsert_payment_records (existing small payments)

Each of these has an insert within them.  Inserts are not idempotent.  So if any of the subsequent process fails, such as upserting the junction object or files, then we will have multiple Bank Account objects being created.

Fairly simple fix.  For progress and spend, break out the code that:
- finds or create bank details
- upserts the junction and files 
Into two separate functions that manage their retries.

Then so something similar for the exist small/M1 payment stuff.

Otherwise, the code seems fine.

PR #1003, opened 2022-06-13
Bug/approved purpose page changes

ptrelease:
## Proposed changes

Fixed linking and error format on approved purposes page

## Further comments

Loops through the collection of approved purposes objects before showing the form.  Populates an href env that links to the offending approved purpose object in the view.

There is no actual reference for this.  So using a counter to get the index of the approved purpose object in the collection.  This is sufficient for linking and should be fine because the ordering of the the in-memory collection is unlikely to change.


Gubbsy (approved):
Looks good

PR #1006, opened 2022-06-13
corrected error message

ptrelease:
## Proposed changes

Corrected error message


ptrelease:
Over shoulder PR done

PR #1011, opened 2022-06-15
Fix/remove old arrears

Gubbsy:
## Proposed changes

Remove old arrears journey


ptrelease (commented):
Gubbsy and I manually tested that arrears, small and M1 payments still work.

ptrelease (approved):

PR #1012, opened 2022-06-15
Ensure errors on add grant cond match design

Gubbsy:
## Proposed changes

Made errors on additional grant conditions match design and error link now goes to first form element (spent some time trying to get the href to link to dynamic checkbox content but could not find an elegant solution within timebox)


ptrelease (approved):
Message appear in correct place

PR #1013, opened 2022-06-20
Feature/update organisation from sf

Gubbsy:
## Proposed changes

Call salesforce to check if an organisation's details need updating and amend the FFE DB if necessary. The update query is performed when the last updated_at date for the Organisation is not equal to the current date, and the organisation is present in SF.


app/helpers/organisation_helper.rb
```diff
@@ -66,4 +69,162 @@ def create_organisation(user)
...
+      else
+
+        Rails.logger.error("Latest organisation details for id: " \
+          "#{organisation.id} populated from salesforce_account_id: " \
+            "#{organisation.salesforce_account_id} failed owing to " \
+              "error: #{e.message}"
+        )
+  
+        # TODO: Raise silent sentry error 
+        # Raven.capture_exception(e)
```
ptrelease, line 143: Given timescales, I'll return and add the usual raise here.

app/helpers/organisation_helper.rb
```diff
@@ -66,4 +69,162 @@ def create_organisation(user)
...
+      #Currently not used as org_types in SF and FFE do not align
+      # organisation.org_type = convert_org_type(sf_details.Organisation_Type__c)
+
+      organisation.save!
+
+      Rails.logger.info("Latest organisation details for id: " \
+        "#{organisation.id} populated from salesforce_account_id: " \
+          "#{organisation.salesforce_account_id}")
+
+    rescue StandardError => e
```
ptrelease, line 116: Checked and Faraday / timeout error inherit from StandardError - so all good here.

app/helpers/organisation_helper.rb
```diff
@@ -66,4 +69,162 @@ def create_organisation(user)
 
   end
 
+  # Method to retrieve and update latest org details from salesforce
+  # aligning FFE with any salesforce changes.
+  #
+  # @param [Organisation] organisation The organisation to update.
+  #
+  def update_existing_organisation_salesforce_details(organisation)
```
ptrelease, line 77: change name of function.  This implies its changing Sakesforce.

app/helpers/organisation_helper.rb
```diff
@@ -66,4 +69,162 @@ def create_organisation(user)
...
+        # TODO: Raise silent sentry error 
+        # Raven.capture_exception(e)
+
+      end
+    end
+  end
+
+  # Method to convert salesforce VAT_registered type to FFE type
+  #
+  # @param [String] vat_registered The string VAT_Register type.
```
ptrelease, line 151: mention its a SF type

app/models/organisation.rb
```diff
@@ -144,6 +151,36 @@ def validate_mission_array
...
+  # continue to pick from option 1-10.
+  #
+  # If options 11, 12 or 13 are used following a Salesforce update - these will
+  # be upserted on new submissions to Salesforce.
+  #
+  # TODO:
+  # - give 11, 12, 13 proper names
+  # - update the functions that set name for upsert to salesforce
+  # - update the functions that get the FFE names from the SF value.
+  # – search the code base for 'org_type' make other changes as required.
```
ptrelease, line 183: update

app/helpers/organisation_helper.rb
ptrelease: On reflection a raise is best.  The small journey has some logic that drives the forms shown based on org type.

So, though a raise may be unnecessary for most applicants, it safeguards against FFE behaving erratically.  

I expect these errors to be pretty uncommon.

PR #1023, opened 2022-06-27
check for bank evidence file before trying to upload

ptrelease:
## Proposed changes

Simple check that the bank details file exists before trying to upload.

Otherwise causes a Sentry alert from the active job if payments exist.  Or an outright rails error if they don't.


Gubbsy (approved):
LG

PR #1026, opened 2022-06-28
Bug/approved purposes errors

ptrelease:
## Proposed changes

Main bug was found in research. Destroy_all was being called against a progress_update_approved_purpose instance which is an error.  Changed to destroy.

During testing, found a second bug where clicking back and trying to update an approved purpose causes an error, because empty approved purposes have since been deleted.

@Gubbsy For now I handle the error by reloading the form.  Not a great user experience because it reloads the form (which is better than an error).  But the user can enter their answer again and its fine.

A better fix would be not deleting the unused approved purposes, but this has knock-ons to other areas of the journey.  Would need more time to fix.


ptrelease:
Includes merge commits - will create a better PR

PR #1029, opened 2022-07-04
Bug/bank account not always uploading

ptrelease:
## Proposed changes

When no bank details are provided, the existing bank details are not given to Salesforce.

When bank details are provided, then a grantee changes their mind to use the old ones, the updated bank details are still used.

This PR solves that by:

 - Unused bank details never go in.
 - Only ask if bank details have changed if there has been a previous arrears payment for that project and the bank details used are not void.
- When we reuse bank details, they should only come from a previous arrears payment for that project, and they should not be void.


app/helpers/progress_and_spend_helper.rb
```diff
@@ -634,9 +634,19 @@ def upload_payment_request(
...
+    if funding_application.payment_details.present?
+
+      # Grantee has provided bank details. Find or create bank account in Salesforce
+      salesforce_bank_account_id =
+        salesforce_api_client.find_or_create_bank_account_details(funding_application)
+
+    else
+
+      # Grantee has not provided new bank details - use what they have supplied already.
+      salesforce_bank_account_id =
```
Gubbsy, line 646: Why are we adding a new query when we have the method `find_or_create_bank_account_details` that returns us a bank account if one already exists? Surely we do not need this conditional statement and new query. We are returned a bank details ID regardless and we can upsert it using `upsert_bank_account_payment_request_junction`

ptrelease: Because find_or_create_bank_account_details queries on account number and sort code that it gets from the payment details.

If a grantee says that the bank details have not change - there are no payment details with which to get the account number and sort code

This query gets the salesforce reference for the latest bank details, and associates to the form.

Gubbsy: Ah that makes sense! I follow now

PR #1043, opened 2022-08-04
Added ordering, default asc, so second signatory guaranteed to get the signatory email

ptrelease:
## Proposed changes

explicit ordering by created_at asc, so that second means created later than first.


Gubbsy (approved):
lg

PR #1049, opened 2022-08-31
Feature/40 m1 pay

Gubbsy:
## Proposed changes

Add 40% journey for M1 applications up to 100k that have already received their 50% payment. 

This is achieved by sending the M1 funding application through the existing arrears payments (similar to large arrears), however, the user is locked to completing this journey once, through status locks. Upon completion, the user can review their payment evidence on the medium task page (again re-using the previous payments page from arrear). 

To achieve this behaviour a new status has been added to indicate that an M1 40 payment can begin and another for the completion of m1 40. A number of context changes and task controller changes have also been made to accommodate the M1 40 re-use of arrears.


app/controllers/concerns/funding_application_context.rb
```diff
@@ -102,12 +103,16 @@ def invalid_view_for_progress_and_spend?
...
         @funding_application&.del_250k_to_5mm?  || \
           (@funding_application&.agreement_submitted_on.present? && \
-              @funding_application&.is_100_to_250k?)
+            (@funding_application&.is_100_to_250k? || @funding_application&.is_10_to_100k?))
 
 
-      payment_can_start = @funding_application&.payment_can_start?
+      progress_spend_payment_allowed =
+        @funding_application&.payment_can_start? || \
+          @funding_application.m1_40_payment_can_start?
```
ptrelease, line 111: @funding_application&.payment_can_start? 

has a safe operator, but 

@funding_application.m1_40_payment_can_start? 

does not.  The safe operator means that a badly pasted url or bookmark will redirect to root - rather than error.

lib/apis/salesforce/salesforce_api.rb
```diff
@@ -2042,6 +2042,62 @@ def get_project_title(salesforce_case_id)
       end
 
     end
+
```
ptrelease, line 2045: Add the rdoc for this

app/controllers/concerns/funding_application_context.rb
ptrelease: added

lib/apis/salesforce/salesforce_api.rb
ptrelease: Added

ptrelease (approved):

ptrelease (changes requested):
Adding VAT form

ptrelease (approved):

PR #1052, opened 2022-09-29
only do payment specific actions when a payment exists

ptrelease:
## Proposed changes

Only store payment request debug info and set the payment amount when a payment is present.

Only remove the debug info a payment request was submitted.

Fixes a bug where project updates only for M2's were failing because payment_request was nil.


Gubbsy (commented):
LG

Gubbsy (approved):

PR #1053, opened 2022-10-05
corrected but that doesnt check flipper before referencing new field

ptrelease:
## Proposed changes

Upsert checks for flipper


Gubbsy (approved):
LG

PR #1054, opened 2022-10-06
Feature/dev40pay

ptrelease:
## Proposed changes

Reused m140pay principals to develop dev40pay journey.  Content and mails were the same, so nearly everything could be reused - just needed to be activated.


ptrelease (commented):
Looks fine.

PR #1057, opened 2022-10-17
Add bank account to permitted routes for progress and spend

Gubbsy:
## Proposed changes

Add bank account to permitted routes for progress and spend to fix big where agreement_submitted_on is not present on older applications.


ptrelease:
Confirmed with Adam that permitting this route does not allow bank details to be provided earlier than appropriate.

PR #1071, opened 2022-12-12
Add fix to threshold validation and change error message

Gubbsy:
Add fix to high spend model to account for spends equal to the threshold value. Amend the error message to reflect this.


config/locales/en.yml
```diff
@@ -969,6 +969,7 @@ en-GB:
             amount:
               not_a_number: "Enter amount of spend"
               greater_than: "Enter a spend greater than £500"
+              greater_than_or_equal_to: "Enter a spend greater than or equal to the spend threshold"
             vat_amount:
```
ptrelease, line 973: Lets discuss if we can use the "Enter a spend greater than £500" content.  And replace £500 with a dynamic string.

Gubbsy: Discussed, and confirmed with Abbie. This content is fine.

ptrelease (approved):

ptrelease:
Waiting on Product decision on whether to allow greater than or equal to.

ptrelease:
Product Owner has signed off on proposed fix.

PR #1086, opened 2023-02-27
reverted code that only provides mission if available

ptrelease:
## Proposed changes

Revert code that removes mission from json on small grant applications if no mission is provided.  The existing Salesforce code does not support this and always requires mission to be supplied. 

A later version of the Salesforce code is ready for release.  At which point we can reintroduce this code.


Gubbsy (approved):
Looks good!

PR #1087, opened 2023-03-02
Feature/large delivery arrears import

Gubbsy:
## Proposed changes

Add capability to import migrated SF4 larger delivery projects from salesforce.

Following on from the already implemented 'is_this_your_project' page, once a user has confirmed the project details are correct. If the user has not previously created an account and the user and organisation details on FFE are blank, then these details are imported from the organisation and contact associated with the project in salesforce, otherwise, if the details are present and align with salesforce then only the project is imported. 

When a project is being imported from salesforce, a new `funding_application` object is created, populated with the required information to progress through an FFE arrears journey, and associated with the current user's organisation. 

Once imported, the user is directed to the project dashboard, and the project is visible under the awarded delivery grants. 
This is achieved through the addition of a new funding_application type `migrated_sf4_large_delivery`, which the dashboard checks against. The application is only visible if it has the status 'payment_can_start', which is set if the permission to start date is present. 

The user can then click the project title from the dashboard, playthrough and submit a progress update/payment request through the existing large arrears journey.


app/controllers/import/is_this_your_project_controller.rb
```diff
@@ -51,24 +51,33 @@ def update
...
 
-      redirect_to :authenticated_root
+      # Orchestrate populating user's org from Salesforce Account if required
+      unless orchestrate_organisation_import_ok?(@salesforce_info_for_page.AccountId)
+        redirect_to organisation_existing_organisations_error_path(
+          current_user.organisations.first.id
+        ) and return
+      end
+
+      orchestrate_funding_application_import_ok?(@salesforce_info_for_page)
```
ptrelease, line 73: Above function does actually return boolean.  But it would be a good idea.  Then we could send throw an error screen and log the error.

ptrelease: Discussed. Made non-boolean

ptrelease (approved):

config/locales/salesforce_text/en.yml
```diff
@@ -45,7 +45,7 @@ en-GB:
       evaluation: "Evaluation"
       contingency: "Contingency"
       new_build_work: "New build work"
-      new_building_work: "New build work" # Salesforce actually uses new_building_work as API name
+      new_building_work: "New building work"
```
ptrelease, line 48: SFF/5 applications not working at the moment - this could be the cause?

ptrelease: Its missing in the SF picklists for large delivery.  So no, this is fine.

PR #1088, opened 2023-03-08
Changed query_all to query to avoid bringing back deleted project costs

ptrelease:
## Proposed changes

Changed query_all to query to avoid bringing back deleted project costs during the legal agreement process.
Also changed query_all code to query when:
- Checking for existing contacts on medium grant submission
- Checking for legal agreement checkbox
- 
As these could easily be checked as part if the same manual testing.


Gubbsy (approved):
LG!

PR #1091, opened 2023-03-14
Bank Account creation and search specifies record Id

ptrelease:
## Proposed changes

Salesforce has a new Bank Account record type that is default.

This will break existing functionality.

This change specifies the record type that should be used and considered by FFE.


Gubbsy (approved):
LG

PR #1097, opened 2023-03-22
Add default record type for bank details

Gubbsy:
## Proposed changes (previously reverted)

Salesforce has a new Bank Account record type that is the default.

This will break existing functionality.

This change specifies the record type that should be used and considered by FFE


ptrelease (approved):
Confirmed the code is he same as what was tested for https://github.com/heritagefund/funding-frontend/pull/1091/files

1091 was reverted because the record types weren't in prod.  Tests passed in lower envs because the record types were present.

PR #1106, opened 2023-05-02
ensured language preferences saved as lowercase on admin portal refresh

ptrelease:
## Proposed changes

Ensured language preferences saved as lowercase on admin portal refresh

Otherwise database will show distinct types as:

english
welsh
both
English
Welsh
Both

Which is poor for data integrity and reporting


Gubbsy (approved):
Looks good

PR #1107, opened 2023-05-03
Add returns after render to avoid double render

Gubbsy:
## Proposed changes

Add returns after redirects and renders on evidence of high spend controller to avoid double render error.


ptrelease (approved):
Looks good.

PR #1113, opened 2023-06-05
Fix/stale arrears journey

Gubbsy:
## Proposed changes

Add check in `invalid_view_for_progress_and_spend?` to ensure that any journeys past the `start` and `select_journey` pages have a valid arrears journey tracker. Otherwise throw the user back to dashboard. 

This will prevent the consistent null arrears_journey_tracker error we see when a user bookmarks or revisits a stale tab of an already submitted arrears journey.


app/controllers/concerns/funding_application_context.rb
```diff
@@ -113,21 +113,38 @@ def invalid_view_for_progress_and_spend?
...
+        (!arrears_tracker_present &&   
+          (request.path.include?('start') || 
+              request.path.include?('select_journey')
+          )
+        ) 
+        has_valid_arrears_tracker = true
+      else
+        has_valid_arrears_tracker = false
+      end
+
```
ptrelease, line 129: Lines 116 to 129 - These are fine, and leave if that's most readable.

You could get the lines down with:

```
no_journey_in_progress =
  @funding_application.arrears_journey_tracker.nil? ||
    (@funding_application.arrears_journey_tracker.nil? && 
      (request.path.include?('start') || request.path.include?('select_journey'))
```

Then use no_journey_in_progress  instead of has_valid_arrears_tracker.  Because has_valid_arrears_tracker is a bit misleading - its nil when you're starting a new payment.  But not a big issue; go with your preference,

ptrelease (approved):

PR #1119, opened 2023-07-13
Rails 7 node modules ruby 3 1 1

Gubbsy:
## Proposed changes

Upgrade to rails 7 includes: 

- Code changes to fix deprecated/broken functionality brought in by the version upgrade. 

- Gem package dependency updates to fix outdated libraries. 

- Update to dependent JS packages.  

Note: Node binaries have been included in the repo to ensure build on gov PaaS, they should be removed and built on platform when moved Azure


.ruby-version
JJD1990, line 1: Gemfile has 3.1.0

app/helpers/application_helper.rb
JJD1990, line 1: does this form builder appear later on?

JJD1990 (dismissed):

app/views/funding_application/gp_open_medium/ownership/show.html.erb
JJD1990, line 1: How does this work? 
<noscript><% no_js = true %></noscript>

app/views/partials/_summary_errors_flash.html.erb
JJD1990, line 1: Is this right? Seems different from the rest of the changes made to errors elsewhere in the PR.

.ruby-version
ptrelease: Used bundle update --ruby top bring the ruby versions up to date without changing the gems we tested with.

app/helpers/application_helper.rb
ptrelease: The require_relative was removed because Rails 7 changed Zeitwork loading.  As a rule, any code that is part of the app is now autoloaded.
Gems still need to be required when used.

app/views/funding_application/gp_open_medium/ownership/show.html.erb
ptrelease: If Javascript is turned off, then the embedded ruby in the <noscript> tags sets the no_js variable (as in no javascript) to true.
Then on line 38, we use no_js to decide whether if remote is true or false.
When remote is true, the request goes directly to the controller function.  When remote is false - Ajax handles the request.

In refreshing my memory on this - I discovered Rails 7 deprecates 'remote' https://stackoverflow.com/questions/20737026/how-does-remote-true-works-in-rails
So I'll leave this unresolved to make any changes and retest.

app/views/partials/_summary_errors_flash.html.erb
ptrelease: Gubbsy spent some time on this.  He could tell you why the original pattern stands.  Essentially he tried the new pattern and it didn't fit with the way the flash error were constructed.  He has not  indented correctly when he put the code back.

ptrelease (approved):

JJD1990 (approved):

etelish (approved):

PR #1127, opened 2023-07-27
test written to mock current_user as nil, & saftey operator added to …

JJD1990:
## Proposed changes

Frequent error seen where the current_user is null as their session has timed out. As a result the funding_application_context organisation check fails with a null reference. A safety selection operator has been added the to the `current_user` to ensure that this null reference is never thrown along with the relevant unit test.


ptrelease:
Interesting problem.  I've manually tested (roughly) what would happen, and the applicant would be redirected to the dashboard which is fine.  But because current_user is still nil, dashboard_controller.show immediately falls down when calling the user_details_complete function which requires current_user.

May be worth considering a defensive check in funding_application_context.rb, so that if current_user is nil, then the user is signed out?  If that works out, then follow up with a test for that?

To test, I set organisation_id to nil on line 21 of funding_application_context.rb.  Then explicitly set the current_user = nil at the start of dashboard_controller.show.

Gubbsy:
> Interesting problem. I've manually tested (roughly) what would happen, and the applicant would be redirected to the dashboard which is fine. But because current_user is still nil, dashboard_controller.show immediately falls down when calling the user_details_complete function which requires current_user.
> 
> May be worth considering a defensive check in funding_application_context.rb, so that if current_user is nil, then the user is signed out? If that works out, then follow up with a test for that?
> 
> To test, I set organisation_id to nil on line 21 of funding_application_context.rb. Then explicitly set the current_user = nil at the start of dashboard_controller.show.

Interesting, I thought the dashboard controller ` before_action :authenticate_user!` would kick in and throw the user back to the login page if it was nill?

PR #1129, opened 2023-08-01
Fix/1280 timed out session null

JJD1990:
## Proposed changes

Frequent error seen where the current_user is null as their session has timed out. As a result the funding_application_context organisation check fails with a null reference. In the progress spend submit_controller.rb there was a before_action: :authenticate_user! which was conflicting with the devise call in the funding_application_context.rb. removed before action.


JJD1990:
We have left comments in the spec as we need to put more work into mocking devise calls. this PR has the bug fix but not the completed tests.

PR #1146, opened 2023-09-12
inital regex alteration to enforce domains in email validation

JJD1990:
## Proposed changes

Altering the regex that is used for email validation to ensure all emails that are entered have at least one domain rather than allowing 'test@test' which does not meet salesforce validation. The regex will also allow emails with tags so 'example+tag@askjeeves.com'


Gubbsy:
Change looks good! Are you able to add a unit test to cover and prove this please? :)

config/initializers/devise.rb
```diff
@@ -171,7 +171,9 @@
   # Email regex used to validate email formats. It simply asserts that
   # one (and only one) @ exists in the given string. This is mainly
```
Gubbsy, line 172: Can you also add to this comment that it also checks that a domain is present?

Gubbsy (approved):
Looks Good

PR #1149, opened 2023-09-22
fix for no users route with relevant tests

JJD1990:
## Proposed changes

Fix in place with relevant tests for the no users route error, card can be seen here:
https://trello.com/c/IDy0cMUn


spec/requests/users/user_request_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+
+    it "allows paths beyond /users" do
+      get "/users/sign_up?"
+      expect(response).to have_http_status(:success)
+    end
+
+  end
+
+end
```
Gubbsy, line 19: Add EOF newline

Gubbsy (approved):

JJD1990:
After research of the workings of devise and having written more tests to check the added route does not interfere with devise the suggested `get /users` route should be ok. More than happy to look into the alternative workaround of changing the back button functionality though if you would prefer? @Gubbsy

Gubbsy (approved):

PR #1151, opened 2023-09-26
Feature/docker containerisation

matthewford:
## Proposed changes

Based on the previous docker work, which runs both the web app and the workers within the same container (more CPU and RAM might need to be added to the container to run both effectively). 

## Further comments

Database migrations are run in the image; but we will also look at using an [init container](https://learn.microsoft.com/en-us/azure/container-instances/container-instances-init-container
) that will run the command for the migrations. However, this will need to be tested; this is preferable to running them within the running container, as we could hit timeouts with a long migration that would kill the container mid-deploy.

## Checklist

- [ ] Test the image on Azure
- [ ] PORT ENV VAR needs to be added in Azure (we might need to change the exposed port depending on what is expected in azure)
- [ ]


Gubbsy:
Merging to test pipelines

PR #1164, opened 2023-10-11
renamed actions and changed to be manually triggered

JJD1990:
## Proposed changes
Have renamed actions so they are easier to tell apart, have retriggered them to be manual.
currently having an issue with deployment to training env (prod, civmuguat and research are not yet ready) @Lubosky please could you look at why the pipeline is failing? https://github.com/heritagefund/funding-frontend/actions/runs/6482891357/job/17603465047


app/assets/stylesheets/application.scss
```diff
@@ -3,7 +3,7 @@
 // ############################################################################
 
 // GOV.UK Settings
-@import "govuk-frontend/govuk/settings/all";
+@import "../../../node_modules/govuk-frontend/govuk/settings/all";
```
Lubosky, line 6: I would avoid using relative paths. If we set load path correctly in `assets.rb` file to be like this `Rails.application.config.assets.paths << Rails.root.join("node_modules")` we can omit this part `../../../node_modules/`

config/initializers/assets.rb
```diff
@@ -6,7 +6,7 @@
 # Add additional assets to the asset load path.
 # Rails.application.config.assets.paths << Emoji.images_path
 # Add Yarn node_modules folder to the asset load path.
-Rails.application.config.assets.paths << Rails.root.join("node_modules")
+Rails.application.config.assets.paths << Rails.root.join("node_modules", "govuk-frontend", "govuk", "assets")
```
Lubosky, line 9: I know we were reverting changes to the previous state but this would only include images and fonts from this specific package in the search path. Probably better to set `node_modules` only as it would make easier to reference packages in the future.

Lubosky (commented):
The issue with the pipelines failing seems to be missing ENV variable so I would not revert all the changes (added some comments why).

The assets are being compiled using production environment so this wasn't really an issue until we added PaperTrail to production. Now it's trying to initialize PaperTrail but it errors out due to missing variable. 

Since I cannot push directly to this branch, I opened PR #1165 with the possible fix as well as cherry-picked commit from this PR renaming the actions to be more meaningful.

JJD1990:
Agree with @Lubosky points, will close this PR and merge his (https://github.com/heritagefund/funding-frontend/pull/1165)

PR #1172, opened 2023-10-24
service unavailable banner added with red border

JJD1990:
## Proposed changes

Added 'service unavailable' banner to notify users that the service will be down for the designated time frame.


Gubbsy (approved):
Looks good!

PR #1174, opened 2023-10-26
inital refactoring and implementaion of new routes, controllers and v…

JJD1990:
…iews for organisation flow

## Proposed changes

For the organisation journey build we need to firstly refactor the page routing and remove the organisation flow from the start of application journeys  and instead, the organisation flow should be presented after the initial user flow.

With this, we can also skeleton templates to the new pages with controllers and views.


config/routes.rb
```diff
@@ -82,37 +82,66 @@
...
+      put '/org-description', to: 'org_description#update'
+      get '/communities-that-org-serve', to: 'communities_that_org_serve#show'
+      put '/communities-that-org-serve', to: 'communities_that_org_serve#update' 
+      get '/leadership-self-identify', to: 'leadership_self_identify#show'
+      put '/leadership-self-identify', to: 'leadership_self_identify#update'
+      get '/charity-number', to: 'charity_number#show'
+      put '/charity-number', to: 'charity_number#update'
+      get '/company-number', to: 'company_number#show'
+      put '/company-number', to: 'company_number#update'
+      get 'vat-registered', to: 'vat_registered#show'
```
Gubbsy, line 99: Can you add the slashes hear for consistency

Gubbsy (approved):

PR #1177, opened 2023-10-31
migration added, new attributes added to org model, new tests for org…

JJD1990:
… spec, some integration tests commented out to get past pipeline, new org types added but commented out

## Proposed changes

For the organisation journey build we need to refactor the existing data model to capture the new question answers and potentially remove data that we no longer require.

- Create migration to add new fields to organisation model, and remove fields to potentially no longer require.

- Refactor organisation model DBO to account for data capture changes - (have done in routes branch)

- Expand on org_type table to account for additional org types captured. (These are entered as enum values rather than table columns) 

- Amend tests, for new data captured.


app/models/organisation.rb
```diff
@@ -146,4 +172,19 @@ def == (other)
       unknown: 11
   }
 
+  # enum new_org_type: {
```
Gubbsy, line 175: We're still going to need this plus our new org types

Gubbsy: Looks good!

Gubbsy (approved):
Looks good to me

PR #1178, opened 2023-11-02
Feature/organisation journey refactor content plus logic

JJD1990:
## Proposed changes

build the actual content for the organisation flow, with required view components and controller logic, validation etc.

Build pages based on design.

Write tests


Gubbsy (approved):
Approved, but I would like to later have a look if we can improve the approach we use for `[board_members_or_trustees_controller.rb](https://github.com/heritagefund/funding-frontend/pull/1178/files#diff-38871d238294636821bf41a0cde78f793123be85d223feab36f303134ea662f7) and the other text within a radio button selections.

PR #1183, opened 2023-11-07
Containerise and prepare application for azure deployment. 

Gubbsy:
## Proposed changes

- Containerise application with docker for deployment on Azure infrastructure. 
- Refactor `database.yaml` to work with containerised application and azure infrastructure. 
- Refactor `storage.yaml` to connect to NLHF owned S3 buckets. 
- Refactor `<env>.rb`s to account for azure deployment. 
- Build new github jobs for Azure deployments. 
- Remove node modules and pre-compile application within docker image. 

## Further comments

If this is a large or complex change, discuss why the solution was chosen and 
alternatives that were considered, etc.

## Checklist


app/assets/stylesheets/components/_cookie-banner.scss
JJD1990, line 1: I remeber there being some confusion around this - if precompilation fails on merge this is a potential area to look at.

config/initializers/assets.rb
JJD1990, line 1: Again, if precompilation fails on merge this is a point to look at.

config/database.yml
JJD1990, line 1: This may need to be looked into in the future, we should be setting the database name on line 9 but we have needed to hardcode for each environment.

Gemfile
JJD1990, line 1: Probably need to deleted the azure-storage-blob lines

JJD1990 (approved):

PR #1186, opened 2023-11-16
Reverted organisation-description to use main_purpose and activities …

Gubbsy:
…as data model.

## Proposed changes

Data insight team have asked that the organisation-description re-uses the existing main_purpose_and_activities model in SF, as a result, FFE should reflect this. This work reverts the organisation description field to use the main_purposes_and_activities model. 

Please rollback and re-run the last migration to update your DB appropriately. 

## Further comments

If this is a large or complex change, discuss why the solution was chosen and 
alternatives that were considered, etc.

## Checklist


JJD1990 (approved):
Looks good!

PR #1191, opened 2023-11-17
check answers logic changed slightly away from not answered to no com…

JJD1990:
…pany number etc locale files altered

## Proposed changes

Have altered the wording in the check answers controller to suit the answer the user has given us. Will be looking at making sure the data persists when a user revisits a previously answered question.


Gubbsy (approved):
Looks good!

PR #1196, opened 2023-11-22
Fix/reddis connection flexibility

Gubbsy:
## Proposed changes

Make reddis connections more flexible in an attempt to fix the Open SSL sentry error we are seeing [https://national-lottery-heritage-fund.sentry.io/issues/4576131777/?environment=production&project=2067916&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=0](here).


JJD1990 (approved):
Makes sense!

PR #1200, opened 2023-11-23
welsh translations added to new org flow, also welsh error msg added …

JJD1990:
## Proposed changes
 Added Welsh translations for new organisation journey


Gubbsy (approved):
Check a few EOF newlines, looks good though

